PR #5, opened 2022-06-21
Add alephium to wormhole

Lbqds:
The wormhole/dev.v2 branch has the latest commit of [wormhole bridge](https://github.com/certusone/wormhole).

You can build a local development network by following the `devnet.md` instructions.

The wormhole contract is located in `alephium/contracts`.
The contract unit tests are located in `alephium/test`.
The guardian watcher is located at `node/pkg/alephium`.


Lbqds:
Unit tests failed because we are using full node `1.4.0-rc3` in `alephium-web3` 0.1.0-rc3 version: https://github.com/alephium/alephium-web3/blob/4082e4dc05c0e812e07b21e6675ba83f43264d4e/package.json#L30

alephium/test/token-bridge-test.ts
```diff
@@ -0,0 +1,610 @@
...
+                    'payer': payer,
+                    'createContractAlphAmount': minimalAlphInContract
+                },
+                inputAssets: [inputAsset],
+                existingContracts: tokenBridgeInfo.dependencies
+            })
+        })
+    })
+
+    it('should create token wrapper for local token', async () => {
```
h0ngcha0, line 123: Is it possible to test that even if the sequence is out of order, it can still be processed as long as its not processed before? (If I understand correctly)

h0ngcha0 (commented):
First round of review, very impressive job!

I have some issues attesting the token for some reasons, since it can not fetch the events from the tx id, which I will look deeper.

alephium/test/governance-test.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    it('should failed if signature is invalid', async () => {
+        const updateGuardianSet = new UpdateGuardianSet(testGuardianSet)
+        const vaaBody = new VAABody(updateGuardianSet.encode(CHAIN_ID_ALEPHIUM), governanceChainId, governanceContractId, 0)
+        const vaa = initGuardianSet.sign(initGuardianSet.quorumSize(), vaaBody)
+        const invalidSignatures = Array(vaa.signatures.length).fill(0).map(_ => randomBytes(66))
+        const invalidVaa = new VAA(vaa.version, vaa.guardianSetIndex, invalidSignatures, vaa.body)
+        await expectOneOfError(
+            async () => await testCase(invalidVaa, 'submitNewGuardianSet'),
+            ["AssertionFailed", "FailedInRecoverEthAddress", "InvalidConversion", "InvalidBytesSliceArg"]
```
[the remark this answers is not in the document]
Lbqds: No, each signature data has 66 bytes and consists of a 1-byte guardian key index and a 65-byte ECDSA signature. Randomly generated data may have invalid guardian key index and ECDSA recovery id, which will result in `InvalidBytesSliceArg` and `InvalidConversion` errors respectively.

alephium/contracts/governance.ral
```diff
@@ -0,0 +1,191 @@
...
+        // check module and governance action
+        assert!(u256From32Byte!(byteVecSlice!(payload, 0, 32)) == coreModule)
+        assert!(byteVecSlice!(payload, 32, 33) == action)
+
+        return payload
+    }
+
+    fn parseAndVerifyGovernanceVAA(vaa: ByteVec, action: ByteVec) -> ByteVec {
+        let coreModule = 0x436f7265 // module identifier: "Core"
+        let payload = parseAndVerifyGovernanceVAAGeneric(vaa, receivedSequence, coreModule, action)
```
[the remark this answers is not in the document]
Lbqds: Wormholes divide messages into governance messages and application messages. Governance messages come from off-chain, and application messages are sent by on-chain contracts. Other wormhole contract implementations avoid replay attacks by storing all message hashes. We avoid replay attacks by validating the sequence.

For governance messages, we need to ensure that the sequence is incremented as it is generated off-chain. For application messages, the on-chain contract guarantees that the sequence is incremented.

alephium/contracts/token_bridge.ral
```diff
@@ -0,0 +1,168 @@
+import "governance.ral"
+import "sequence.ral"
```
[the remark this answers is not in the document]
Lbqds: Forgot to remove it, thanks!

alephium/contracts/sequence.ral
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    fn checkSequence(seq: U256) -> () {
+        if (seq < next) {
+            let undoneSequenceId = subContractId!(u256To8Byte!(seq / 256))
+            UndoneSequence(undoneSequenceId).check(seq)
+            return
+        }
+
+        if (seq >= (next + 512)) {
```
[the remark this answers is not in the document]
Lbqds: Sounds great 👍

alephium/contracts/token_wrapper.ral
```diff
@@ -0,0 +1,100 @@
...
+        assert!(msgTokenId == tokenContractId)
+        assert!(msgTokenChainId == getTokenChainId())
+
+        let totalAmount = deNormalizeAmount(amount, decimals_)
+        let transferFee = deNormalizeAmount(arbiterFee, decimals_)
+        assert!(totalAmount > transferFee)
+        let transferAmount = totalAmount - transferFee
+        let recipient = byteVecToAddress!(#00 ++ toAddress)
+        let tokenId = getTokenId()
+        transferAlph!(caller, recipient, dustAmount!())
```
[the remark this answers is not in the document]
Lbqds: Yes 👍

alephium/test/token-bridge-test.ts
Lbqds: The test case is here: https://github.com/alephium/wormhole-fork/pull/5/files#diff-48dda3fbdbf4f3b4cdc3e31e6c1b1d3583b5e9dbaf658415b81c4f57db974263R134, I will add more sequence related test cases to the token bridge.

alephium/test/governance-test.ts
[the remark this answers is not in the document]
h0ngcha0: Got it, thanks!

alephium/test/token-bridge-test.ts
h0ngcha0: Awesome, thanks!

alephium/contracts/governance.ral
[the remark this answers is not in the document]
h0ngcha0: So as part of getting Alephium into an existing Wormhole network, we basically negotiate with all the guardians and agree that the sequence should be incremented correctly. Got it.

Thanks for the detailed explanation!

alephium/contracts/token_bridge/wrapped_alph.ral
```diff
@@ -0,0 +1,34 @@
+Contract WrappedAlph(
+    mut totalWrapped: U256
+) {
+    @using(preapprovedAssets=true, assetsInContract=true, permissionCheck=false)
+    pub fn wrapAndTransfer(
+        fromAddress: Address,
+        totalAmount: U256
+    ) -> () {
+        totalWrapped = totalWrapped + totalAmount
```
h0ngcha0, line 9: Do we want to check the caller of this method?

alephium/contracts/token_bridge/wrapped_alph.ral
```diff
@@ -0,0 +1,34 @@
...
+
+    @using(preapprovedAssets=true, assetsInContract=true, permissionCheck=false)
+    pub fn unwrapAndCompleteTransfer(
+        fromAddress: Address,
+        recipient: Address,
+        transferAmount: U256,
+        arbiter: Address,
+        arbiterFee: U256
+    ) -> () {
+        let totalAmount = transferAmount + arbiterFee
```
h0ngcha0, line 25: Do we want to check the caller of this method?

alephium/contracts/governance.ral
```diff
@@ -0,0 +1,202 @@
...
+    const CoreModule = 0x436f7265 // module identifier: "Core"
+
+    enum ActionId {
+        ContractUpgrade = #01
+        NewGuardianSet  = #02
+        NewMessageFee   = #03
+        TransferFee     = #04
+    }
+
+    @using(readonly = true, permissionCheck = false)
```
h0ngcha0, line 24: Do we need to have `permissionCheck = false` here?

alephium/contracts/governance.ral
```diff
@@ -0,0 +1,202 @@
...
+        let (targetChainId, payload) = parseAndVerifyGovernanceVAA(vaa, ActionId.NewGuardianSet)
+        assert!(targetChainId == chainId || targetChainId == 0, ErrorCodes.InvalidTargetChainId)
+        let newGuardianSetIndex = u256From4Byte!(byteVecSlice!(payload, 33, 37))
+        assert!(newGuardianSetIndex == guardianSetIndexes[1] + 1, ErrorCodes.InvalidGuardianSetIndex)
+        let newGuardianSetSize = u256From1Byte!(byteVecSlice!(payload, 37, 38))
+        assert!(newGuardianSetSize > 0, ErrorCodes.InvalidGuardianSetSize)
+        let payloadSize = 38 + newGuardianSetSize * 20
+        assert!(size!(payload) == payloadSize, ErrorCodes.InvalidMessageSize)
+        guardianSetIndexes[1] = newGuardianSetIndex
+        guardianSets[1] = byteVecSlice!(payload, 37, payloadSize)
```
h0ngcha0, line 182: Should this be `guardianSets[1] = byteVecSlice!(payload, 38, payloadSize)` ?

h0ngcha0 (commented):
Impressive work :rocket: 
I have some questions & comments

alephium/contracts/sequence/sequence.ral
```diff
@@ -0,0 +1,62 @@
+// sequence started from 0, set `next = 0` when init contract state
+Abstract Contract Sequence(
+    mut next: U256,
+    mut next1: U256, // TODO: better name
+    mut next2: U256,
+    undoneSequenceTemplateId: ByteVec,
+    refundAddress: Address
+) extends Constants() {
+    @using(readonly = true)
+    fn setExecuted(offset: U256, current: U256) -> U256 {
```
h0ngcha0, line 10: [Nit] If this method is `setExecuted` perhaps `UndoneSequence` can be named `UnExecutedSequence`? Or perhaps `UnfinishedSequence`. 

Undone initially gives me the impression that it is gonna un-do an operation :)

alephium/contracts/sequence/sequence.ral
```diff
@@ -0,0 +1,62 @@
+// sequence started from 0, set `next = 0` when init contract state
+Abstract Contract Sequence(
+    mut next: U256,
+    mut next1: U256, // TODO: better name
+    mut next2: U256,
```
h0ngcha0, line 5: Question: Instead of using `next1` and `next2` here, if we use one `sequence` field just like in the `UndoneSequence` contract, is the benefit that we will create less `UndoneSequence` sub contract?

alephium/contracts/token_bridge/token_bridge_for_chain.ral
```diff
@@ -0,0 +1,69 @@
...
+        tokenPool.completeTransfer{caller -> dustAmount!()}(
+            remoteChainId, amount, vaaTokenId, vaaTokenChainId, recipient, normalizedArbiterFee, caller
+        )
+    }
+
+    pub fn destroyUndoneSequenceContracts(paths: ByteVec) -> () {
+        checkPermission!(callerContractId!() == localTokenBridgeId, ErrorCodes.InvalidCaller)
+        let length = size!(paths)
+        let mut index = 0
+        while (index < length) {
```
h0ngcha0, line 63: Question: what is the use case for this command?

If this is executed successfully, what is the consequence for the `Sequence` contract? Would the `checkSequence` method still work?

alephium/contracts/sequence/undone_sequence.ral
```diff
@@ -0,0 +1,26 @@
+Contract UndoneSequence(
+    parentId: ByteVec,
+    begin: U256,
+    mut sequences: U256,
+    refundAddress: Address
+) extends Constants() {
+    @using(assetsInContract = true)
+    pub fn checkUndone(seq: U256) -> () {
```
h0ngcha0, line 8: [Nit] perhaps `setExecuted` or `checkSequence` ?

alephium/contracts/token_bridge/token_bridge_for_chain.ral
```diff
@@ -0,0 +1,69 @@
...
+        assert!(size!(payload) == 131, ErrorCodes.InvalidMessageSize)
+        let recipient = byteVecToAddress!(#00 ++ toAddress)
+        return amount, vaaTokenId, vaaTokenChainId, recipient, normalizedArbiterFee
+    }
+
+    @using(preapprovedAssets = true)
+    pub fn completeTransfer(vaa: ByteVec, caller: Address) -> () {
+        let (sequenceReady, payload) = checkTransfer(vaa)
+        if (!sequenceReady) { // too many undone sequences in between, compact those sequences first
+            return
```
h0ngcha0, line 47: Question: just curious, for the relayer, how does it know when to re-try?

alephium/contracts/sequence/sequence.ral
```diff
@@ -0,0 +1,62 @@
...
+    @using(assetsInContract = true)
+    fn compact() -> () {
+        if (next1 != SequenceExecutedFlag) {
+            let encodedState = encodeToByteVec!(
+                selfContractId!(),
+                next,
+                next1,
+                refundAddress
+            )
+            copyCreateSubContract!{selfAddress!() -> 1 alph}(u256To8Byte!(next / 256), undoneSequenceTemplateId, encodedState)
```
h0ngcha0, line 25: Question: Would it be possible that an attacker from the other chain flood the chain with small transfer messages (small amount), drains the ALPH balance from the `Sequence` and  therefore disable it? I guess someone would then need to call the `deposit` method to top up the balance?

alephium/contracts/sequence/undone_sequence.ral
```diff
@@ -0,0 +1,26 @@
+Contract UndoneSequence(
+    parentId: ByteVec,
+    begin: U256,
+    mut sequences: U256,
+    refundAddress: Address
```
h0ngcha0, line 5: Would it make sense to have `refundAddress` being the address of the `Sequence` contract (or its subclass)? Since `Sequence` contract is the one that spent the ALPH and created the `UndoneSequence`, I guess if the money goes back to it when all sequences in `UndoneSequence` are checked, it will reduce the amount of manual intervention.

alephium/contracts/token_bridge/remote_token_pool.ral
```diff
@@ -0,0 +1,66 @@
...
+            callerContractId!(), emitterChainId, amount, vaaTokenId, vaaTokenChainId, normalizedArbiterFee
+        )
+
+        if (caller != recipient) {
+            transferAlph!(caller, recipient, dustAmount!())
+        }
+        let tokenId = selfContractId!()
+        transferTokenFromSelf!(recipient, tokenId, transferAmount)
+        if (arbiterFee > 0) {
+            transferTokenFromSelf!(caller, tokenId, arbiterFee)
```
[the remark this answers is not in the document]
Lbqds: There is no relayer at the beginning, and users need to submit transactions by themselves, but in the future, I think a part of `messageFee` can be used as an incentive.

alephium/contracts/token_bridge/token_bridge.ral
```diff
@@ -0,0 +1,249 @@
...
+
+    @using(preapprovedAssets = true)
+    pub fn attestToken(payer: Address, localTokenId: ByteVec, nonce: ByteVec, consistencyLevel: U256) -> () {
+        assert!(size!(localTokenId) == 32, ErrorCodes.InvalidTokenId)
+        assert!(size!(nonce) == 4, ErrorCodes.InvalidNonceSize)
+        if (localTokenId != wrappedAlphId) {
+            assert!(tokenRemaining!(payer, localTokenId) > 0, ErrorCodes.TokenNotExist) // make sure that the token exists
+        }
+
+        // We don't set decimals, symbol and name for local token due to sharding
```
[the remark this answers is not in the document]
Lbqds: Because users may create tokens on other groups, and we can't get these informations by contract call.

alephium/contracts/token_bridge/attest_token_handler.ral
```diff
@@ -0,0 +1,37 @@
...
+  remoteChainId: U256,
+  remoteTokenBridgeId: ByteVec,
+  mut receivedSequence: U256
+) extends Constants(), TokenBridgeConstants() {
+  @using(preapprovedAssets = true)
+  pub fn handleAttestToken(
+    vaa: ByteVec,
+    payer: Address,
+    createContractAlphAmount: U256
+  ) -> () {
```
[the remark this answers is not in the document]
Lbqds: Sounds great 👍 , I'll improve this later.

alephium/contracts/token_bridge/token_bridge_for_chain.ral
```diff
@@ -0,0 +1,69 @@
...
+        return checkSequence(msgSequence), payload
+    }
+
+    @using(readonly = true)
+    fn parseCompleteTransfer(payload: ByteVec) -> (U256, ByteVec, U256, Address, U256) {
+        assert!(byteVecSlice!(payload, 0, 1) == PayloadId.Transfer, ErrorCodes.InvalidPayloadId)
+        let amount = u256From32Byte!(byteVecSlice!(payload, 1, 33))
+        let vaaTokenId = byteVecSlice!(payload, 33, 65)
+        let vaaTokenChainId = u256From2Byte!(byteVecSlice!(payload, 65, 67))
+        let toAddress = byteVecSlice!(payload, 67, 99)
```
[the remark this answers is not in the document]
Lbqds: This is because we changed the wormhole protocol. In the wormhole protocol, the messages sent by the token bridge of each chain to other chains will share the same sequence, which makes it difficult for us to deal with the sequence(because we use sequence instead of message hash to avoid replay attack). In the wormhole protocol, `toChainId` usually appears in the payload, we put `toChainId` in the header and changed ETH contracts in [this PR](https://github.com/alephium/wormhole-fork/pull/10), but our documentation has not been updated, there is [an issue](https://github.com/alephium/wormhole-fork/issues/12) to track this.

alephium/contracts/token_bridge/attest_token_handler.ral
```diff
@@ -0,0 +1,37 @@
+Contract AttestTokenHandler(
+  governanceContractId: ByteVec,
+  localChainId: U256,
+  localTokenBridgeId: ByteVec,
```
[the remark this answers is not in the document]
Lbqds: I was hesitant about this before because ETH uses `contract address` and we use `contract id`.

alephium/contracts/token_bridge/attest_token_handler.ral
```diff
@@ -0,0 +1,37 @@
...
+    vaa: ByteVec,
+    payer: Address,
+    createContractAlphAmount: U256
+  ) -> () {
+    let governance = Governance(governanceContractId)
+    let (emitterChainId, targetChainId, emitterAddress, msgSequence, payload) = governance.parseAndVerifyVAA(vaa)
+    assert!(emitterChainId == remoteChainId, ErrorCodes.InvalidEmitChainId)
+    assert!(emitterAddress == remoteTokenBridgeId, ErrorCodes.InvalidEmitAddress)
+    assert!(targetChainId == localChainId || targetChainId == 0, ErrorCodes.InvalidTargetChainId)
+    assert!(msgSequence >= receivedSequence, ErrorCodes.InvalidSequence)
```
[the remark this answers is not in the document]
Lbqds: 1. No, the `AttestToken` message is sent by the contract on the chain, and the sequence will increase by 1 every time a message is sent. For example, ff the user wants to attest a token on ETH to Alephium, the user needs to call the ETH contract, and it will publish the `AttestToken` message, but If the user did not complete the attestation with the `AttestToken` message, then the message may become invalid later, and user need to re-do attestation.
2. No, the contract only needs to ensure that every time an `AttestToken` message is sent, the sequence will be incremented by one.

alephium/devnet/deploy.ts
```diff
@@ -0,0 +1,108 @@
...
+        consts.payer,
+        parseInt(process.env.INIT_CHAIN_ID!),
+        parseInt(process.env.INIT_GOV_CHAIN_ID!),
+        process.env.INIT_GOV_CONTRACT!,
+        initGuardianSet,
+        0,
+        consts.messageFee
+    )
+
+    const contracts = await wormhole.deployContracts("devnet")
```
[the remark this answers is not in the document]
Lbqds: I've also thought about this before, it's technically possible to deploy to multiple groups, but if we deploy to multiple groups, we also need to:

1. Deploy wrapped ALPH on multiple groups
2. Deploy the token wrapper on multiple groups
3. Listen for token bridge messages from multiple groups

And I'm not sure if it will affect the user experience, but maybe we can give it a try.

alephium/contracts/token_bridge_scripts/transfer_local.ral
```diff
@@ -0,0 +1,29 @@
+TxScript Main(
+    tokenBridgeId: ByteVec,
+    fromAddress: Address,
+    localTokenId: ByteVec,
+    alphChainId: U256,
+    toChainId: U256,
+    toAddress: ByteVec,
+    tokenAmount: U256,
+    messageFee: U256,
```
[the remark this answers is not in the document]
Lbqds: My initial thought was that we should rarely update the `messageFee`, this script is called by the bridge UI, so we only need to update the bridge UI's config every time we update the `messageFee`, which can reduce the gas fee of the transaction, but maybe I'm overthinking.

alephium/contracts/token_bridge/token_bridge.ral
```diff
@@ -0,0 +1,249 @@
...
+    pub fn transferAlph(
+        fromAddress: Address,
+        toChainId: U256,
+        toAddress: ByteVec,
+        alphAmount: U256,
+        arbiterFee: U256,
+        nonce: ByteVec,
+        consistencyLevel: U256
+    ) -> () {
+        transferToken{fromAddress -> alphRemaining!(fromAddress)}(fromAddress, wrappedAlphId, localChainId, toChainId, toAddress, alphAmount, arbiterFee, nonce, consistencyLevel)
```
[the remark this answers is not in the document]
Lbqds: Same as above, this reduces one function call.

alephium/lib/wormhole.ts
```diff
@@ -0,0 +1,384 @@
...
+        this.refundAddress = refundAddress
+        this.localChainId = localChainId
+        this.governanceChainId = governanceChainId
+        this.governanceEmitterAddress = governanceEmitterAddress
+        this.initGuardianSet = initGuardianSet
+        this.initGuardianSetIndex = initGuardianSetIndex
+        this.initMessageFee = initMessageFee
+    }
+
+    private async remotTokenPoolContract(): Promise<Contract> {
```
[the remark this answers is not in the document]
Lbqds: Thanks!

alephium/contracts/token_bridge/local_token_pool.ral
```diff
@@ -0,0 +1,48 @@
...
+        recipient: Address,
+        normalizedArbiterFee: U256,
+        caller: Address
+    ) -> () {
+        let (transferAmount, arbiterFee) = prepareCompleteTransfer(
+            callerContractId!(), emitterChainId, amount, vaaTokenId, vaaTokenChainId, normalizedArbiterFee
+        )
+
+        if (caller != recipient) {
+            transferAlph!(caller, recipient, dustAmount!())
```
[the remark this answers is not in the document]
Lbqds: There are two possible cases:
1. the `caller` is the recipient, then we don't need to transfer `dustAmount` of ALPH
2. the `caller` is not the recipient, and we only have approved assets from the `caller`, we need to transfer `dustAmount` of ALPH to the recipient, so the output of the recipient is `dustAmount ALPH + transferred token`

alephium/contracts/token_bridge/wrapped_alph.ral
Lbqds: It would be great if we can check the caller, but I don't have a good way to do this 😓

alephium/contracts/token_bridge/wrapped_alph.ral
Lbqds: Same as above.

alephium/contracts/governance.ral
Lbqds: Yes, `permissionCheck` is true for public functions by default, we need to make it false to avoid compiler warnings.

alephium/contracts/governance.ral
Lbqds: The encoding of `guardianSet` is: 1 + 20 * (the number of guardians), the first byte is the number of guardians, so we use 37 here.

alephium/lib/wormhole.ts
```diff
@@ -0,0 +1,384 @@
...
+        }
+    }
+
+    private async deployGovernance(
+        networkType: NetworkType,
+        governanceChainId: number,
+        governanceEmitterAddress: string,
+        initGuardianSet: string[],
+        initGuardianSetIndex: number,
+        initMessageFee: bigint
```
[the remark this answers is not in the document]
Lbqds: Yes, thanks!

alephium/contracts/sequence/sequence.ral
Lbqds: Sounds good to me 👍

alephium/contracts/sequence/sequence.ral
Lbqds: Yes, there is another benefit, the transaction execution order in the block may be inconsistent with the sequence order, if there is only one field in the `Sequence` contract, it may create an unnecessary `UndoneSequence` contract, and using two fields can greatly reduce the probability of this problem.

alephium/contracts/token_bridge/token_bridge_for_chain.ral
Lbqds: `UndoneSequence` contract should not or rarely be created in the normal case, but if we are attacked, it may create lots of `UndoneSequence` contracts, so we add a governance action to delete these contracts.

alephium/contracts/sequence/undone_sequence.ral
Lbqds: Sounds good, thanks!

alephium/contracts/token_bridge/token_bridge_for_chain.ral
Lbqds: The relayer can check the `Sequence` contract state. If the relayer has a message in which the sequence is `seq`, the relayer can re-try if `sequence.next + 512 > seq`.

alephium/contracts/sequence/sequence.ral
Lbqds: Yes, if this happens we need to remove these `UndoneSequence` contracts by the `DestroyUndoneSequenceContracts` governance action.

alephium/contracts/sequence/undone_sequence.ral
Lbqds: It would be great if the `refundAddress` is the `Sequence` contract, but the `destroySelf` only supports asset addresses.

Lbqds:
Thank you very much for your great review ❤️ @h0ngcha0

alephium/contracts/token_bridge/remote_token_pool.ral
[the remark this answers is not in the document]
h0ngcha0: Yeah, agree. The fee should probably be denominated as the native token of the originating chain. But that is for the future I guess.

alephium/contracts/token_bridge/attest_token_handler.ral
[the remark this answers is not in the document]
h0ngcha0: Sounds great :+1:

alephium/contracts/token_bridge/token_bridge_for_chain.ral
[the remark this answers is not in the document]
h0ngcha0: I see, thanks for the great explanation!

alephium/contracts/token_bridge/attest_token_handler.ral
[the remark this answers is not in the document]
h0ngcha0: Sorry, in this case it should be called id since it is a "contract id", my bad.

alephium/contracts/token_bridge/attest_token_handler.ral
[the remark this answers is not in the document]
h0ngcha0: Thanks for the explanation :+1:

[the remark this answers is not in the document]
h0ngcha0: I understand why we use sequence to prevent replay attack. I guess compared to using hash, the cons is that we put higher requirement on the correctness of the smart contract on the other chain, since it now has the power to break our handler if there is a bug on sequencing on the their side, whereas with the hash approach, that is not possible.

alephium/devnet/deploy.ts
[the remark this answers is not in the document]
h0ngcha0: Thanks. If we do not deploy them to multiple groups, what does that mean to UX?

alephium/contracts/token_bridge/attest_token_handler.ral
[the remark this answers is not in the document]
Lbqds: Yes, for other chain implementations, even if the sequence is the same, the verification can be successful as long as the message hash is different, but the sequence cannot be the same for us.

alephium/devnet/deploy.ts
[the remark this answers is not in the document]
Lbqds: Sorry, I thought about it again, and feels like we can completely hide these details, which should have no impact on the user experience.

alephium/contracts/token_bridge_scripts/transfer_local.ral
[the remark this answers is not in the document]
h0ngcha0: That makes sense, consider this as a [Nit] :)

alephium/contracts/token_bridge/local_token_pool.ral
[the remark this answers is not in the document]
h0ngcha0: Thanks for the explanation, I remember now that the dust amount is per output.

alephium/contracts/token_bridge/wrapped_alph.ral
h0ngcha0: Good point. Perhaps we can check the `callerCodeHash`? I guess we know the code hash of the `WrappedAlphPool` when we create `WrappedAlph` contract.

Since we only have one copy of `WrappedAlphPool`, maybe that is enough? Otherwise it feels a bit dangerous since the attacker can use an arbitrary contract id to initiate the `WrappedAlphPool` here.

h0ngcha0: Hmm, maybe even the above is not enough...

Do we need to actually have a field in `WrappedAlph` to remember the contract id of the `WrappedAlphPool` after `WrappedAlphPool` is created, and only allow these two methods to be called when the contract id is set?

h0ngcha0: Ok, maybe it is ok, since we restricted the amount of tokens that can be received here

```
WrappedAlphPool(callerContractId!()).receiveWrappedAlph{
   self -> selfTokenId!(): totalAmount
}(self, totalAmount)
```

Lbqds: I'm thinking that maybe we can merge `WrappedAlph` to `WrappedAlphPool` if the `WrapperAlph` is only used for the bridge. Because we only have one `WrappedAlph` and one `WrappedAlphPool` instance.

alephium/contracts/token_bridge/token_bridge.ral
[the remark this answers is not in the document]
polarker: Even if the data is available on chain, it's not trustworthy because attackers could create another contract with the same token information.

alephium/contracts/token_bridge/wrapped_alph.ral
h0ngcha0: Yeah, I thought about that too, I kinda like the fact that `WrapperAlph` makes everything conceptually the same: a "token". But perhaps it is a good idea to merge it :thinking:

alephium/contracts/token_bridge/token_bridge_for_chain.ral
h0ngcha0: I guess `sequence.next` grows when relayer retries, right? If it tries enough times it will probably make `sequence.next + 512 > seq`?

alephium/contracts/token_bridge/attest_token_handler.ral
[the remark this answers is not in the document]
polarker: I think sequence should be unique by design and by code:
1. The sequence is updated when publishing message, so every new message (bridge action) would have different sequence number.
2. Sequence seems designed to be unique:
```
        // Nonce of the VAA, must to be set to random bytes. Nonces
	// prevent collisions where one emitter publishes identical
	// messages within one block (= timestamp).
	//
	// It is not suitable as a global identifier -
	// use the (chain, emitter, sequence) tuple instead.
	Nonce uint32 // <-- NEW
```
3. A lot of wormhole's code relies on sequence number. If sequence cannot be used as UUID, then it will be a problem for them too.

alephium/contracts/sequence/sequence.ral
h0ngcha0: Would that make the legitimate transfers during the attack invalid? Tokens are deposited correctly on the other side, but we wont be able to withdraw it from ALPH.

alephium/contracts/governance.ral
h0ngcha0: Awesome, thanks for the explanation :+1:

alephium/contracts/token_bridge/token_bridge_for_chain.ral
h0ngcha0: Yeah, got it. I was thinking about the legitimate ones, like in this comment: https://github.com/alephium/wormhole-fork/pull/5#discussion_r958259375

So basically all the "good" ones during the attack will be invalid, which could be an attack if the attacker knows that there is a huge tranfer (10M dollars)  and just flood the system with very small transfers to sabotage that.

alephium/contracts/token_bridge/token_bridge_for_chain.ral
Lbqds: The sequence grows only if the transaction is successful. Therefore, if `seq > sequence.next + 512` , the relayer can regularly query the sequence contract state. When `sequence.next + 512 > seq`, the message will be executed successfully.

alephium/contracts/sequence/sequence.ral
Lbqds: The `UndoneSequence` contract is used to store those messages that have not been withdrawn in time so that the sequence contract will not be blocked, if the normal user withdraws in time, then the `UndoneSequence` only have the sequences from attack messages. So it depends on when we destroy these undone sequence contracts.

alephium/contracts/token_bridge/token_bridge_for_chain.ral
Lbqds: In this case, we have to wait for the normal user to execute withdraw before destroy the undone sequence contracts.

alephium/contracts/sequence/sequence.ral
polarker, line 5: How about naming these 3 as `start`, `firstNext256`, `secondNext256`?

alephium/contracts/token_bridge/token_bridge.ral
```diff
@@ -0,0 +1,247 @@
...
+            0,
+            0
+        )
+        let path = Path.TokenPool ++ u256To2Byte!(localChainId) ++ wrappedAlphId
+        let wrappedAlphPoolTemplateId = TokenBridgeFactory(tokenBridgeFactoryId).getWrappedAlphPoolTemplateId()
+        copyCreateSubContract!{payer -> createContractAlphAmount}(path, wrappedAlphPoolTemplateId, encodedState)
+    }
+
+    @using(preapprovedAssets = true)
+    pub fn createLocalTokenPool(localTokenId: ByteVec, payer: Address, createContractAlphAmount: U256) -> () {
```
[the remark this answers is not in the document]
Lbqds: I'll improve this, thanks!

alephium/contracts/token_bridge/attest_token_handler.ral
```diff
@@ -0,0 +1,37 @@
...
+  pub fn handleAttestToken(
+    vaa: ByteVec,
+    payer: Address,
+    createContractAlphAmount: U256
+  ) -> () {
+    let governance = Governance(governanceContractId)
+    let (emitterChainId, targetChainId, emitterAddress, msgSequence, payload) = governance.parseAndVerifyVAA(vaa)
+    assert!(emitterChainId == remoteChainId, ErrorCodes.InvalidEmitChainId)
+    assert!(emitterAddress == remoteTokenBridgeId, ErrorCodes.InvalidEmitAddress)
+    assert!(targetChainId == localChainId || targetChainId == 0, ErrorCodes.InvalidTargetChainId)
```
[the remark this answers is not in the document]
Lbqds: I also found this problem, the `targetChainId` should always be 0.

alephium/contracts/sequence/sequence.ral
Lbqds: Sounds good 👍

polarker:
I have done my last big round of review, all looks great to me now. There are two more minor ideas:

We could decompose token_bridge.ral into an abstract contract token_bridge_governance.ral and token_bridge.ral which inherits the abstract governance contract. This would be similar to Ethereum and probably help code review.
We could add a test for public functions for the main contracts like token_bridge.ral, so we make sure that we will not break the visibilities of the functions.
I will also spend sometime to think about how to improve WrappedAlph and refundAddress.

polarker:
For the rest of the week, I will review the non-contract part of the repo.

polarker (approved):
LGTM, a few minor comments left.

Dockerfile.const
```diff
@@ -41,3 +42,4 @@ COPY --from=const-build /scripts/.env.0x ethereum/.env
 COPY --from=const-build /scripts/.env.hex solana/.env
 COPY --from=const-build /scripts/.env.hex terra/tools/.env
 COPY --from=const-build /scripts/.env.hex algorand/.env
+COPY --from=const-build /scripts/.env.alph alephium/.env
```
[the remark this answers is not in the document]
Lbqds: When creating `RegisterChain` VAA, the sequences for other chains are randomly generated (from legacy code), for us the sequences are incremental, so I put them in a separate file, the code is here: https://github.com/alephium/wormhole-fork/blob/50e7696010f8efbffa6e22ce6483d94545dfe623/scripts/guardian-set-init.sh#L11

alephium/devnet/consts.ts
```diff
@@ -0,0 +1,8 @@
+export const payer = '1DrDyTr9RpRsQnDnXo2YRiPzPW4ooHX5LLoqXrqfMrpQH'
+export const messageFee = BigInt("100000000000000")
+export const consistencyLevel = 0
+export const oneAlph = BigInt("1000000000000000000")
+export const minimalAlphInContract = oneAlph
+export const testWalletName = 'alephium-web3-test-only-wallet'
+export const testWalletPassword = 'alph'
+export const testWalletMnemonic = 'vault alarm sad mass witness property virus style good flower rice alpha viable evidence run glare pretty scout evil judge enroll refuse another lava'
```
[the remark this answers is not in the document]
Lbqds: The code is outdated, we have removed the `devent` directory in the `improve-deployment-workflow` branch.

alephium/lib/wormhole.ts
```diff
@@ -0,0 +1,397 @@
...
+    Fields,
+    addressFromContractId,
+    subContractId,
+    Project
+} from '@alephium/web3'
+import { waitTxConfirmed, zeroPad } from './utils'
+
+const Byte32Zero = "0".repeat(64)
+const DummyRefundAddress = addressFromContractId(Byte32Zero)
+const MaxALPHValue = BigInt("1000000000") * BigInt("1000000000000000000")
```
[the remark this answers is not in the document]
Lbqds: Thanks!

polarker:
Merged to master as this PR is way too large!

PR #10, opened 2022-07-05
Modify message id and improve sequence contract

Lbqds:
* In the wormhole protocol, messages are uniquely identified by `emitterChainId + emitterContractAddress + sequence`, which means that the sequence of messages sent to Alephium from other chains is discontinuous. In order to verify the message more conveniently, we need to modify the message id to `emitterChainId + emitterContractAddress + targetChainId + sequence`
* Improve the sequence contract to avoid DoS attack


alephium/contracts/sequence.ral
```diff
@@ -30,11 +30,16 @@ TxContract Sequence(
...
-            return
+            return true
+        }
+
+        if (seq >= (next + 768)) {
+            compact()
+            return false
         }
 
         if (seq >= (next + 512)) {
```
polarker, line 45: Do we need a return for this branch?

Lbqds: No, in this case, we will move the first window to the undone sequence contract and continue to verify the `seq`.

polarker: I see! `next` is updated in `compact`

node/pkg/db/db.go
```diff
@@ -161,7 +170,8 @@ func (d *Database) FindEmitterSequenceGap(prefix VAAID) (resp []uint64, firstSeq
 		}
 
 		// Find min/max (yay lack of Go generics)
-		first := false
+		// TODO: add tests
+		first := true
```
polarker: Why do we need to change this now?

node/cmd/guardiand/adminclient.go
```diff
@@ -184,9 +190,6 @@ func runFindMissingMessages(cmd *cobra.Command, args []string) {
 	for _, id := range resp.MissingMessages {
 		fmt.Println(id)
 	}
-
-	log.Printf("processed %s sequences %d to %d (%d gaps)",
-		emitterAddress, resp.FirstSequence, resp.LastSequence, len(resp.MissingMessages))
```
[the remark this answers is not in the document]
Lbqds: I remove it by accident, thanks!

node/cmd/guardiand/adminserver.go
```diff
@@ -90,16 +90,11 @@ func adminContractUpgradeToVAA(req *nodev1.ContractUpgrade, timestamp time.Time,
 		return nil, errors.New("invalid new_contract address")
 	}
 
-	if req.ChainId > math.MaxUint16 {
```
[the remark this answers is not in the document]
Lbqds: It has been checked here: https://github.com/alephium/wormhole-fork/pull/10/files#diff-cbb74a4fbdf9c64a7c79e6796c768552d4aeb3af65b356b9867d91fea7d6ceb9R170

node/pkg/db/db.go
Lbqds: This function is used to get lost messages from other guardians, I misunderstood it at first, I reverted and added some tests in a later commit: https://github.com/alephium/wormhole-fork/pull/10/commits/9853146687a108c023f13c102668f45b237fde70

polarker:
LGTM, huge amount of great work 👍  I did not review the Ralph contract changes, which I will do it different.

polarker (approved):

PR #11, opened 2022-07-05
Improve token transfer

polarker:


alephium/contracts/token_bridge.ral
```diff
@@ -103,21 +102,16 @@ TxContract TokenBridge(
     pub fn attestToken(payer: Address, tokenId: ByteVec, nonce: ByteVec, consistencyLevel: U256) -> () {
         assert!(size!(tokenId) == 32)
         assert!(size!(nonce) == 4)
+        // FIXME: tokenExists!(tokenId)?
+        // assert!(tokenRemaining!(tokenId) != 0) // make sure that the token exists
```
Lbqds, line 106: Do we need to add a built-in function to check if the contract exists?

polarker: 1. We cannot check the contract of tokenId here, because the token might be created in other group.
2. In general, I was hesitating about such a function. It might be useful to have it for subcontracts, but we don't need it in our project so far.

WDYT?

polarker: `assert!(tokenRemaining!(tokenId) != 0)` is enough to check the existence of the contract?

Lbqds: Indeed, we cannot check the tokens created in other groups.

The `tokenRemaining` has two parameters: `address` and `tokenId`, does it means the caller need to approve some token when calling this method?

polarker: Yes, need to approve. We could even check it `== 1` here, and always approve `1`.

Lbqds: Got it, thanks.

PR #18, opened 2022-08-09
More updates

Lbqds:
1. Check the minimal confirmations in the contract
2. Support the following case(I tested with ALPH, ETH, and BSC in my local devnet environment.):

   > The token bridges guarantee that there will be a unique wrapped asset on each chain for each non-native token. In other words, transferring a native token from chain A to chain C will result in the same wrapped token as transferring from A to B first, then from B to C, and no double wrapping will happen.

3. Improve the bridge UI


Lbqds:
The test will fail because the field `tokenBridgeForChainId` in contract `*TokenPool` is unused. But keeping this field is helpful for debugging. After we release the new sdk version, the test will pass.

alephium/devnet/deploy.ts
```diff
@@ -32,24 +32,22 @@ async function createWallet() {
...
-    tokenBridgeForChainId: string,
-    remoteChain: string
+    tokenBridgeId: string
 ) {
-    let txId = await wormhole.createLocalTokenPool(tokenBridgeForChainId, localTokenId, consts.payer, consts.minimalAlphInContract)
+    let txId = await wormhole.createLocalTokenPool(tokenBridgeId, localTokenId, consts.payer, consts.minimalAlphInContract)
     let tokenPoolAddress = await getCreatedContractAddress(provider, txId, 0)
     const tokenPoolId = binToHex(contractIdFromAddress(tokenPoolAddress))
-    console.log('token pool id for ' + remoteChain + ': ' + tokenPoolId)
+    console.log('token pool id: ' + tokenPoolId)
```
polarker, line 40: Do we also want to log the token id here?

polarker (commented):
LGTM, a few minor comments

alephium/contracts/token_bridge/token_bridge.ral
```diff
@@ -151,8 +152,9 @@ Contract TokenBridge(
         tokenAmount: U256,
         arbiterFee: U256,
         nonce: ByteVec,
-        consistencyLevel: U256 // TODO: Do we need to limit the minimum confirmation
+        consistencyLevel: U256
```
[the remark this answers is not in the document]
Lbqds: Yes, `consistencyLevel` is fixed on the ETH contract, does it make sense to allow users to set the `consistencyLevel` if they want to transfer a large amount of token?

alephium/contracts/token_bridge/token_bridge.ral
```diff
@@ -22,6 +22,7 @@ Contract TokenBridge(
     refundAddress: Address
 ) extends Constants(), TokenBridgeConstants() {
     const TokenBridgeModule = 0x546f6b656e427269646765 // module identifier: "TokenBridge"
+    const MinimalConsistencyLevel = 10
```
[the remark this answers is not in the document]
Lbqds: It's fixed in the ETH contract, if we allow the user to set the `consistencyLevel`, we need to make sure `consistencyLevel > MinmalConsistencyLevel` to avoid problems caused by reorg.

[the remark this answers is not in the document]
Lbqds: Maybe 10 is a bit too much, users need to wait about 10 minutes

alephium/contracts/token_bridge/token_bridge.ral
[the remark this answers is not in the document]
polarker: Agree, we'd better to make this configurable

Lbqds:
I'll update the bridge UI in another PR.

sdk/js/src/utils/consts.ts
```diff
@@ -611,3 +611,7 @@ function evm_chain_subset(e: EVMChainName): ChainName {
   // will fail to compile if 'e' can't be typed as a [[ChainName]]
   return e;
 }
+
+export const MAINNET_ALPH_MINIMAL_CONSISTENCY_LEVEL = 105
+export const TESTNET_ALPH_MINIMAL_CONSISTENCY_LEVEL = 105
```
polarker, line 616: Let's use a smaller number for testnet? maybe `10` as you initially set.

alephium/lib/wormhole.ts
```diff
@@ -27,13 +22,13 @@ interface NetworkConfigs {
 
 const networkConfigs: Record<NetworkType, NetworkConfigs> = {
     "mainnet": {
-        minimalConsistencyLevel: MAINNET_ALPH_MINIMAL_CONSISTENCY_LEVEL
```
polarker: Why do we make this change? Isn't it good to reuse the constants?

sdk/js/src/utils/consts.ts
Lbqds: Done

alephium/lib/wormhole.ts
Lbqds: Because we have not released the SDK, CI can't download the dependency, I plan to release the SDK after the contract is finalized, and I'll update this once we release the SDK. WDYT?

polarker: agreed

PR #20, opened 2022-08-17
Add transaction page to bridge ui

Lbqds:
For security reasons, we need to wait for more than 100 confirmations for a transfer tx, so we need to save the transaction history so that users can easily redeem when the transaction is confirmed. The idea is:

1. save the transaction data to the browser [indexeddb](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) when the user creates transactions
2. fetch the transaction data from the explorer if the browser cache is cleared


h0ngcha0:
A higher level question before I dig deeper into the go code :) 
Why do we need to store txs  in the txDB in the workhole node?

bridge_ui/src/components/Transactions.tsx
```diff
@@ -0,0 +1,350 @@
...
+}
+
+async function updateTxDB(
+  nodeProvider: NodeProvider,
+  fromAddress: string,
+  targetChainId: ChainId,
+  isTransferCompleted: (sequence: string) => Promise<boolean>
+) {
+  const count = await transactionDB.txs.count()
+  if (count === 0) {
```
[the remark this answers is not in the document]
Lbqds: Sorry, I don't quite understand what you mean, if there are transactions in the DB, then the count should not be 0.

bridge_ui/src/components/Transactions.tsx
```diff
@@ -0,0 +1,350 @@
...
+async function updateTxStatus(nodeProvider: NodeProvider, targetChainId: ChainId) {
+  const pendingTxs = await getTxsByStatus("Pending", CHAIN_ID_ALEPHIUM, targetChainId)
+  const pendingTxsStatus = await Promise.all(
+    pendingTxs.map((tx) => nodeProvider.transactions.getTransactionsStatus({txId: tx.txId}))
+  )
+  const removedTxs: string[] = []
+  const confirmedTxs: Transaction[] = []
+  pendingTxs.forEach((tx, index) => {
+    const txStatus = pendingTxsStatus[index]
+    if (isAlphTxNotFound(txStatus)) {
```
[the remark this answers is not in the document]
Lbqds: Yes, I'll fix this, thanks!

bridge_ui/src/components/Transactions.tsx
```diff
@@ -0,0 +1,350 @@
...
+  const [loadingError, setLoadingError] = useState<any>(null)
+
+  useEffect(() => {
+    if (alphSigner && ethProvider) {
+      const update = async () => {
+        const isTransferCompleted = async (sequence: string): Promise<boolean> => {
+          const response = await getSignedVAAWithRetry(
+            CHAIN_ID_ALEPHIUM,
+            ALEPHIUM_TOKEN_BRIDGE_CONTRACT_ID,
+            CHAIN_ID_ETH,
```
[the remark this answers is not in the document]
Lbqds: It's a good idea if we want to show transactions from other chains, but I'm not sure if we want to show transactions from other chains.

bridge_ui/src/components/Transactions.tsx
```diff
@@ -0,0 +1,350 @@
...
+                color="inherit"
+                className={classes.loader}
+              />
+            ) : loadingError ? (
+              <Alert severity="error" className={classes.error}>
+                Loading transactions error: {loadingError}
+              </Alert>
+            ) : (
+              <>
+                <ListTransaction
```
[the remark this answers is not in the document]
Lbqds: Yes, it would be great if we show the "Completed" transactions, I'll improve the transactions page UI later.

Lbqds:
> A higher level question before I dig deeper into the go code :) Why do we need to store txs in the txDB in the workhole node?

The tx DB will be removed if the user clears the browser history, there are two ways to retrieve user transaction information:

1. we have no event filter system like ETH, so we need to get all events emitted by the bridge contract first, then filter out the transactions sent by the user according to the address
2. save the transaction information to the guardian node backend, and fetch the transaction info from the backend if the browser cache is cleared

I think the second method is simpler and more efficient, WDYT?

bridge_ui/src/components/Transactions.tsx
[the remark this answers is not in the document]
h0ngcha0: Pls disregard this comment, I saw the code where new txs are put into the txDb.

h0ngcha0:
> > A higher level question before I dig deeper into the go code :) Why do we need to store txs in the txDB in the workhole node?
> 
> The tx DB will be removed if the user clears the browser history, there are two ways to retrieve user transaction information:
> 
> 1. we have no event filter system like ETH, so we need to get all events emitted by the bridge contract first, then filter out the transactions sent by the user according to the address
> 2. save the transaction information to the guardian node backend, and fetch the transaction info from the backend if the browser cache is cleared
> 
> I think the second method is simpler and more efficient, WDYT?

Thanks for the explanation. Do you think using explorer is an option? Can get get enough information if it is an option?

h0ngcha0:
> > > A higher level question before I dig deeper into the go code :) Why do we need to store txs in the txDB in the workhole node?
> > 
> > 
> > The tx DB will be removed if the user clears the browser history, there are two ways to retrieve user transaction information:
> > 
> > 1. we have no event filter system like ETH, so we need to get all events emitted by the bridge contract first, then filter out the transactions sent by the user according to the address
> > 2. save the transaction information to the guardian node backend, and fetch the transaction info from the backend if the browser cache is cleared
> > 
> > I think the second method is simpler and more efficient, WDYT?
> 
> Thanks for the explanation. Do you think using explorer is an option? Can we get enough information if it is an option?

Or perhaps that introduces an extra undesirable trust assumption?

Lbqds:
> > > A higher level question before I dig deeper into the go code :) Why do we need to store txs in the txDB in the workhole node?
> > 
> > 
> > The tx DB will be removed if the user clears the browser history, there are two ways to retrieve user transaction information:
> > 
> > 1. we have no event filter system like ETH, so we need to get all events emitted by the bridge contract first, then filter out the transactions sent by the user according to the address
> > 2. save the transaction information to the guardian node backend, and fetch the transaction info from the backend if the browser cache is cleared
> > 
> > I think the second method is simpler and more efficient, WDYT?
> 
> Thanks for the explanation. Do you think using explorer is an option? Can we get enough information if it is an option?

Do you mean the explorer-backend? I thought about this before, as far as I know, our explorer-backend currently does not support filtering transactions based on the calling contract address, so we still need to filter out bridge transactions from all user transactions, there are two ways to check if a transaction is a bridge transaction,

1. decode the `scriptOpt` field of the transaction
2. calling the `getEventByTxId`, and check if the event is the bridge transfer event

I feel both are not very efficient, and we can use the explorer backend if it supports filtering transactions based on the calling contract address in the future.

h0ngcha0:
> > > > A higher level question before I dig deeper into the go code :) Why do we need to store txs in the txDB in the workhole node?
> > > 
> > > 
> > > The tx DB will be removed if the user clears the browser history, there are two ways to retrieve user transaction information:
> > > 
> > > 1. we have no event filter system like ETH, so we need to get all events emitted by the bridge contract first, then filter out the transactions sent by the user according to the address
> > > 2. save the transaction information to the guardian node backend, and fetch the transaction info from the backend if the browser cache is cleared
> > > 
> > > I think the second method is simpler and more efficient, WDYT?
> > 
> > 
> > Thanks for the explanation. Do you think using explorer is an option? Can we get enough information if it is an option?
> 
> Do you mean the explorer-backend? I thought about this before, as far as I know, our explorer-backend currently does not support filtering transactions based on the calling contract address, so we still need to filter out bridge transactions from all user transactions, there are two ways to check if a transaction is a bridge transaction,
> 
> 1. decode the `scriptOpt` field of the transaction
> 2. calling the `getEventByTxId`, and check if the event is the bridge transfer event
> 
> I feel both are not very efficient, and we can use the explorer backend if it supports filtering transactions based on the calling contract address in the future.

Practically, it might not be that much of an issue,

h0ngcha0:
> [uting guidelines](/alephium/wormhole-fork/blob/85bbf51035a518326de73ff7a4c364215f39b5ad/CONTRIBUTING.md).

Yeah, i mean explorer-backend.

I

h0ngcha0:
Sorry, accidentally closed the PR :sweat: Re-opened again.

h0ngcha0:
> > > > A higher level question before I dig deeper into the go code :) Why do we need to store txs in the txDB in the workhole node?
> > > 
> > > 
> > > The tx DB will be removed if the user clears the browser history, there are two ways to retrieve user transaction information:
> > > 
> > > 1. we have no event filter system like ETH, so we need to get all events emitted by the bridge contract first, then filter out the transactions sent by the user according to the address
> > > 2. save the transaction information to the guardian node backend, and fetch the transaction info from the backend if the browser cache is cleared
> > > 
> > > I think the second method is simpler and more efficient, WDYT?
> > 
> > 
> > Thanks for the explanation. Do you think using explorer is an option? Can we get enough information if it is an option?
> 
> Do you mean the explorer-backend? I thought about this before, as far as I know, our explorer-backend currently does not support filtering transactions based on the calling contract address, so we still need to filter out bridge transactions from all user transactions, there are two ways to check if a transaction is a bridge transaction,
> 
> 1. decode the `scriptOpt` field of the transaction
> 2. calling the `getEventByTxId`, and check if the event is the bridge transfer event
> 
> I feel both are not very efficient, and we can use the explorer backend if it supports filtering transactions based on the calling contract address in the future.

I think I get what you are saying about the efficiency, thanks. I feel a little bit uncertain because if we could leverage explorer, it would reduce the code & operation complexity here, and I am not sure practically how much we would suffer in terms of efficiency, especially if we are smart about the UI.

I assume most of the time we will probably be interested in transactions that are pretty new, it is probably rare that we redeem something that is transferred days earlier.  Maybe we could only show the recent transactions related to bridge address and  give user the option to see earlier ones. Explorer should have pagination for fetching user transactions already.

WDYT?

Lbqds:
> > > > > A higher level question before I dig deeper into the go code :) Why do we need to store txs in the txDB in the workhole node?
> > > > 
> > > > 
> > > > The tx DB will be removed if the user clears the browser history, there are two ways to retrieve user transaction information:
> > > > 
> > > > 1. we have no event filter system like ETH, so we need to get all events emitted by the bridge contract first, then filter out the transactions sent by the user according to the address
> > > > 2. save the transaction information to the guardian node backend, and fetch the transaction info from the backend if the browser cache is cleared
> > > > 
> > > > I think the second method is simpler and more efficient, WDYT?
> > > 
> > > 
> > > Thanks for the explanation. Do you think using explorer is an option? Can we get enough information if it is an option?
> > 
> > 
> > Do you mean the explorer-backend? I thought about this before, as far as I know, our explorer-backend currently does not support filtering transactions based on the calling contract address, so we still need to filter out bridge transactions from all user transactions, there are two ways to check if a transaction is a bridge transaction,
> > 
> > 1. decode the `scriptOpt` field of the transaction
> > 2. calling the `getEventByTxId`, and check if the event is the bridge transfer event
> > 
> > I feel both are not very efficient, and we can use the explorer backend if it supports filtering transactions based on the calling contract address in the future.
> 
> I think I get what you are saying about the efficiency, thanks. I feel a little bit uncertain because if we could leverage explorer, it would reduce the code & operation complexity here, and I am not sure practically how much we would suffer/gain in terms of efficiency, especially if we are smart about the UI.
> 
> I assume most of the time we will probably be interested in transactions that are pretty new, it is probably rare that we redeem something that was transferred days earlier. Maybe we could only show the recent transactions related to bridge address and give user the option to see earlier ones. Explorer should have pagination for fetching user transactions already.
> 
> WDYT?

Sounds great 👍, I'll remove the txDB from the guardian node and try to use the explorer later, thanks!

h0ngcha0:
> > > > > > A higher level question before I dig deeper into the go code :) Why do we need to store txs in the txDB in the workhole node?
> > > > > 
> > > > > 
> > > > > The tx DB will be removed if the user clears the browser history, there are two ways to retrieve user transaction information:
> > > > > 
> > > > > 1. we have no event filter system like ETH, so we need to get all events emitted by the bridge contract first, then filter out the transactions sent by the user according to the address
> > > > > 2. save the transaction information to the guardian node backend, and fetch the transaction info from the backend if the browser cache is cleared
> > > > > 
> > > > > I think the second method is simpler and more efficient, WDYT?
> > > > 
> > > > 
> > > > Thanks for the explanation. Do you think using explorer is an option? Can we get enough information if it is an option?
> > > 
> > > 
> > > Do you mean the explorer-backend? I thought about this before, as far as I know, our explorer-backend currently does not support filtering transactions based on the calling contract address, so we still need to filter out bridge transactions from all user transactions, there are two ways to check if a transaction is a bridge transaction,
> > > 
> > > 1. decode the `scriptOpt` field of the transaction
> > > 2. calling the `getEventByTxId`, and check if the event is the bridge transfer event
> > > 
> > > I feel both are not very efficient, and we can use the explorer backend if it supports filtering transactions based on the calling contract address in the future.
> > 
> > 
> > I think I get what you are saying about the efficiency, thanks. I feel a little bit uncertain because if we could leverage explorer, it would reduce the code & operation complexity here, and I am not sure practically how much we would suffer/gain in terms of efficiency, especially if we are smart about the UI.
> > I assume most of the time we will probably be interested in transactions that are pretty new, it is probably rare that we redeem something that was transferred days earlier. Maybe we could only show the recent transactions related to bridge address and give user the option to see earlier ones. Explorer should have pagination for fetching user transactions already.
> > WDYT?
> 
> Sounds great +1, I'll remove the txDB from the guardian node and try to use the explorer later, thanks!

Thanks! :pray: :heart_decoration:

bridge_ui/src/components/Transactions.tsx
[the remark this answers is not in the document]
Lbqds: Sorry, I misunderstood this, we don't need to set the "Completed" status here because the data source is the indexeddb. This means that these confirmed transactions have not been redeemed, which will be marked as "Completed" after redeeming, and we only need to update the pending transactions.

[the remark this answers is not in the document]
h0ngcha0: Got it, thanks! :+1:

h0ngcha0 (approved):
LGTM :muscle:

bridge_ui/src/hooks/useHandleTransfer.tsx
```diff
@@ -393,7 +395,16 @@ async function alephium(
         return result.txId
       }
     )
+    await TransactionDB.getInstance().txs.put(new Transaction(
+      txInfo.txId,
+      signer.account.address,
+      CHAIN_ID_ALEPHIUM,
+      targetChain,
+      txInfo.sequence,
+      "Pending"
```
polarker, line 404: Do we also want to cache information of the transfer tx, like amount, toAddress, .etc?

bridge_ui/src/utils/consts.ts
```diff
@@ -291,6 +291,13 @@ export const ALEPHIUM_HOST =
     ? "http://localhost:12973"
     : "http://localhost:22973"
 
+export const ALEPHIUM_EXPLORER_HOST =
+  CLUSTER === "mainnet"
+    ? "http://localhost:9090"
+    : CLUSTER === "testnet"
+    ? "http://localhost:9090"
+    : "http://localhost:9090"
```
polarker, line 299: We could use the same configuration as here: https://github.com/alephium/desktop-wallet/blob/5f10be35b12cef027c47876230652c57a5453a73/src/utils/settings.ts

bridge_ui/src/components/Transactions.tsx
```diff
@@ -0,0 +1,519 @@
...
+  tx: explorer.Transaction
+): Promise<TxIdAndSequence | undefined> {
+  if (typeof tx.inputs === 'undefined' || tx.inputs.length === 0) {
+    return undefined
+  }
+  const haveContractInput = tx.inputs.some(input => typeof input.unlockScript === 'undefined')
+  if (!haveContractInput) {
+    return undefined
+  }
+  const events = await nodeProvider.events.getEventsTxIdTxid(tx.hash)
```
polarker, line 100: I created a Github issued to support events on explorer-backend.

bridge_ui/src/components/Transactions.tsx
```diff
@@ -0,0 +1,519 @@
...
+        getTokenBridgeAddressForChain(CHAIN_ID_ETH),
+        ethProvider,
+        response.vaaBytes
+      )
+    }
+    throw Error("Only support ETH as target chain")
+  }, [ethProvider, txTargetChain])
+
+  // load more transactions from explorer
+  const loadMoreTxsCallback = useCallback(async () => {
```
polarker, line 312: I will think how to make this more efficient.

Does Wormhole's UI uses Etherscan for Tx history?

polarker (approved):
Great work as usual! Minor comments

bridge_ui/src/hooks/useHandleTransfer.tsx
Lbqds: I'm hesitant about this. 

If we save the tx info just for redemption, I guess these fields are not necessary because the UI will show these fields when the user redeems(parse from the VAA). 

But if we want users to be able to find all transaction information more easily, then it would be great to cache these fields.

bridge_ui/src/utils/consts.ts
Lbqds: It's for consistent with the legacy code, all configs use this style in the bridge UI code, and there are lots of codes like `CLUSTER === "mainnet"`.

bridge_ui/src/components/Transactions.tsx
Lbqds: Thanks!

bridge_ui/src/components/Transactions.tsx
Lbqds: No, the wormhole bridge UI won't support tx history, users need to search from Etherscan by themself.

bridge_ui/src/utils/consts.ts
polarker: Alright then!

PR #25, opened 2022-08-30
Minor updates

polarker:


polarker:
@Lbqds I need your help to fix the tests. I don't have the time for it.

Lbqds (approved):
LGTM, thanks!

alephium/contracts/token_bridge/token_bridge_for_chain.ral
```diff
@@ -62,7 +62,6 @@ Contract TokenBridgeForChain(
         for (let mut index = 0; index < length; index = index + 8) {
             let undoneSequenceId = subContractId!(byteVecSlice!(paths, index, index + 8))
             UndoneSequence(undoneSequenceId).destroy()
-            index = index + 8
```
polarker, line 66: Thanks!

h0ngcha0 (approved):
LGTM :+1:

polarker:
I have done my last big round of review, all looks great to me now. There are two more minor ideas:
1. We could decompose `token_bridge.ral` into an abstract contract `token_bridge_governance.ral` and `token_bridge.ral` which inherits the abstract governance contract. This would be similar to Ethereum and probably help code review.
2. We could add a test for public functions for the main contracts like `token_bridge.ral`, so we make sure that we will not break the visibilities of the functions.

I will also spend sometime to think about how to improve WrappedAlph and refundAddress.

PR #27, opened 2022-09-02
Improve deployment workflow

Lbqds:
Inspired by hardhat and truffle, divide the deployment workflow into multiple deploy scripts, feedback is welcome.


alephium/configuration.ts
```diff
@@ -0,0 +1,59 @@
...
+        "scripts/get_test_token.ts",
+        "scripts/create_test_token_pool.ts",
+        "scripts/create_wrapped_alph_pool.ts"
+      ]
+    },
+
+    "mainnet": {
+      // TODO: update config
+      providerUrl: "http://localhost:22973",
+      mnemonic: "vault alarm sad mass witness property virus style good flower rice alpha viable evidence run glare pretty scout evil judge enroll refuse another lava",
```
polarker, line 46: This would be loaded from environment variable I guess.

polarker (commented):
I read through it quickly, really like this approach. Many of the code/design can be reused on `web3` later ❤️

@h0ngcha0 I remember you used a similar approach for the NFT project, your comment on the design would be valuable.

h0ngcha0:
> I read through it quickly, really like this approach. Many of the code/design can be reused on `web3` later heart
> 
> @h0ngcha0 I remember you used a similar approach for the NFT project, your comment on the design would be valuable.
> 
> Will review in details once the PR is mature.

Definitely! :100:

h0ngcha0:
I love the approach as well. I think we can benefit a lot from keeping the DevX similiar to hardhat and other popular tools <3

alephium/configuration.ts
```diff
@@ -0,0 +1,61 @@
...
+const configuration: Configuration = {
+  sourcePath: "contracts",
+  artifactPath: "artifacts",
+
+  networks: {
+    "devnet": {
+      nodeUrl: "http://localhost:22973",
+      mnemonic: "vault alarm sad mass witness property virus style good flower rice alpha viable evidence run glare pretty scout evil judge enroll refuse another lava",
+      deploymentFile: ".deployments.json",
+      scripts: [
```
h0ngcha0, line 12: How does the DevX compare to by default running all scripts in "script" directory in the alphabetic order? Hardhat deploy seems to behave this way, which makes some to adopt the following naming convention for deployment scripts: "00_xxx.js", "01_yyy.js", etc. 

e.g. https://github.com/smartcontractkit/hardhat-starter-kit/tree/main/deploy

alephium/lib/deployment.ts
```diff
@@ -0,0 +1,248 @@
...
+    return value
+  }
+
+  return {
+    provider: signer.provider,
+    account: account,
+    deployContract: deployContract,
+    runScript: runScript,
+    setEnvironment: setEnvironment,
+    getEnvironment: getEnvironment
```
h0ngcha0, line 175: For deployment, I am wondering if it is a good pattern to have multiple scripts running separately and use `setEnvironment` & `getEnvironment` to pass information across the scripts, or is it better to just have a "deploy" script that deploys all contracts :thinking: 

Reference: https://stackoverflow.com/questions/69812460/how-to-deploy-multiple-smart-contracts-using-hardhat-deploy

alephium/deploy.ts
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env node
+import yargs from 'yargs/yargs';
+import { hideBin } from "yargs/helpers"
+
+import configuration from "./configuration"
+import { deploy, NetworkType } from "./lib/deployment"
+
+const argv = yargs(hideBin(process.argv))
```
h0ngcha0, line 8: Nice :+1: 

Not related to this PR. If we wanna share this code across different projects, would it make sense to have a separate CLI for this? just like hardhat, brownie or [dapp.tools](https://github.com/dapphub/dapptools/blob/master/src/dapp/README.md#basic-usage-a-tutorial)? i guess other common tasks are: `compile`, `test`, `node`,  `deploy`, etc

alephium/configuration.ts
h0ngcha0: I realized that the scripts are different between devnet and mainnet, which is a good reason to specify all the scripts here. but this question still holds: https://github.com/alephium/wormhole-fork/pull/27/files#r963145258

Lbqds: We can also use the same way as hardhat/truffle, because the `DeployFunction` have a `NetworkType` parameter, users can decide whether to deploy the contract according to the network type, which way do you prefer?

alephium/lib/deployment.ts
Lbqds: Both hardhat and truffle support multiple scripts running separately, I guess the benefit is to continue execution from the last failed step, the truffle will deploy a `Migration` contract which is used to record successful steps, and for the hardhat deploy plugin, this information will be saved in `.migration.json`.

But yes, the `setEnvironment` & `getEnvironment` is not friendly for developers, truffle/hardhat will save the contract deploy result once the `deployer.deploy()` execute successfully, so the developer doesn't need to call the `setEnvironment`, I'm working on improvements to support this.

alephium/deploy.ts
Lbqds: Yes, I think we also need to support these commands like truffle/hardhat. I'll add a separate CLI later.

alephium/lib/deployment.ts
Lbqds: I pushed a few commits, now the DevX is similar to hardhat and truffle, users can get the contract deployment results by ContractId:

1. hardhat examples: https://github.com/wighawag/hardhat-deploy#deploy-scripts-tags-and-dependencies
2. truffle examples: https://github.com/wormhole-foundation/wormhole/blob/dev.v2/ethereum/migrations/3_deploy_bridge.js

But if the contract is created by another contract:

1. for `createSubContract`, we can get the contract id by `subContractId`
2. for `createContract`, we need to get the transaction output, I think we can add some utility functions to SDK to make it easier

alephium/deploy.ts
h0ngcha0: Nice :+1: 

Where do you plan to put the CLI? Should it be part of the web3 package or a separate repo? I guess the benefit of putting it in the same repo as web3 is that devs can be sure that the CLI version is the same as what the user specified in `package.json`, perhaps by using npx. WDYT?

Lbqds: Yes, it should be part of the alephium-web3, users can use `npx alephium deploy/test/etc` like hardhat/truffle, I'll test with bridge first, and when it's ready I'll add it to the SDK.

h0ngcha0: Awesome :+1:

alephium/lib/deployment.ts
```diff
@@ -0,0 +1,284 @@
...
+  network: Network,
+  deployContractResults: Map<string, DeployContractResult>,
+  runScriptResults: Map<string, RunScriptResult>
+): Deployer {
+  const signer = PrivateKeySigner.from(network.mnemonic)
+  const account = signer.getAccountSync()
+
+  const deployContract = async (contract: Contract, params: DeployContractParams): Promise<DeployContractResult> => {
+    const result = await contract.transactionForDeployment(signer, params)
+    await signer.submitTransaction(result.unsignedTx, result.txId)
```
h0ngcha0, line 163: Do you think it makes sense to deploy only when contract code is changed? We probably can make the assumption that the deployment steps are supposed to be idempotent, therefore can be run multiple times.

I guess that is another way to solve the problem where the whole deployment failed in the middle. Seems to be hardhat-deploy's philosophy:

```
In any case, as a general advice every deploy function should be idempotent. This is so they can always recover from failure or pending transaction. This is what underpin most of hardhat-deploy philosophy.
```

alephium/lib/deployment.ts
```diff
@@ -150,6 +155,49 @@ export class PrivateKeySigner extends SignerWithNodeProvider {
...
+  }
+  for (const key of leftKeys) {
+    if (left[key] !== right[key]) {
+      return false
+    }
+  }
+  return true
+}
+
+async function needToDeploy(
```
h0ngcha0, line 177: Great change :+1: 

Question: With this change (which makes it idempotent), do we still need to check the [lastFailedStep](https://github.com/alephium/wormhole-fork/blob/improve-deployment-workflow/alephium/lib/deployment.ts#L348)? If the user actually change e.g. the code after a failed deployment, does it pick it up?

Lbqds: > With this change (which makes it idempotent), do we still need to check the [lastFailedStep](https://github.com/alephium/wormhole-fork/blob/improve-deployment-workflow/alephium/lib/deployment.ts#L348)?

I was also hesitant about this, if we check the `lastFailedSetp`, then we can skip the successfully executed scripts, which can avoid some extra checks.

> If the user actually change e.g. the code after a failed deployment, does it pick it up?

Yes, it will retry to deploy after a failed deployment regardless of whether the code changes, but if the deployment succeeds and the code changes, then it will skip re-deploy the contract, from this point of view, maybe we should skip to checking the `lastFailedStep`? WDYT?

h0ngcha0: Yeah, it feels like if we expect that the deployment steps should be idempotent (and we have done our part to make it easy to do so) then it feels like we don't need to check the `lastFailedStep` any more. WDYT?

Lbqds: Sounds great 👍 , I'll fix this.

Lbqds: I found another problem, the script will fail if it is executed repeatedly, so we also need to skip the scripts which executed successfully before.

But because we use `Script.typeId` as the task key and users may run the script more than once([register_chain](https://github.com/alephium/wormhole-fork/pull/27/files#diff-4b96891847f33c9f9874491068010d6735e8918d0deba756a43f0c04b0cac88c) is an example), this means different tasks have same task key, so I add an extra optional params `taskName` to `Deployer.runScript`, users don't need this if one script is a task, I'm not sure if there is a better way to resolve this, do you have any suggestions?

alephium/lib/deployment.ts
```diff
@@ -0,0 +1,389 @@
...
+      return runScriptResult
+    }
+    // we have checked in `needToRunScript`
+    return previous!
+  }
+
+  const getDeployContractResult = (typeId: string): DeployContractResult => {
+    const result = deployContractResults.get(typeId)
+    if (result === undefined) {
+      throw new Error(`Contract ${typeId} deployment result does not exist`)
```
polarker, line 303: [Nit]: Deployment result of contract "${typeId}" does not exist

alephium/lib/deployment.ts
```diff
@@ -0,0 +1,389 @@
...
+    }
+    return result
+  }
+
+  const getRunScriptResult = (typeId: string): RunScriptResult => {
+    const result = runScriptResults.get(typeId)
+    if (result === undefined) {
+      throw new Error(`Script ${typeId} execute result does not exist`)
```
polarker, line 311: [Nit]: Execution result of script "${typeId}" does not exist

alephium/contracts/devnet/deploy_governance.ral
```diff
@@ -1,4 +1,4 @@
-TxScript Main(
+TxScript DevnetDeployGovernance(
```
polarker, line 1: Why don't we use the same TxScripts to deploy to testnet and mainnet? Is it for testing purpose?

Since we now persist the deployment results, we might be able to remove these TxScripts.

alephium/lib/deployment.ts
```diff
@@ -0,0 +1,389 @@
...
+  fromGroup: number
+  toGroup: number
+  txId: string
+  blockHash: string
+  codeHash: string
+  attoAlphAmount?: string
+  tokens?: Record<string, string>
+}
+
+export interface DeployContractResult extends ExecutionResult {
```
polarker, line 52: Remark: we will check what fields are missing compared to Hardhat's DeployContractResult. At a first glance, Hardhat has gas information.

Could create Github issues for this.

polarker (commented):
Done my first round of review. Great work 🚀

alephium/configuration.ts
Lbqds: Sounds great 👍 , will update.

alephium/configuration.ts
```diff
@@ -0,0 +1,61 @@
...
+        "scripts/deploy_governance.ts",
+        "scripts/deploy_wrapped_alph.ts",
+        "scripts/deploy_token_bridge_factory.ts",
+        "scripts/deploy_token_bridge.ts",
+        "scripts/register_chain.ts",
+        "scripts/create_wrapped_alph_pool.ts"
+      ]
+    },
```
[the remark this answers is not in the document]
Lbqds: Sounds great 👍 , will update.

alephium/lib/deployment.ts
```diff
@@ -0,0 +1,389 @@
...
+}
+
+type RunScriptParams = Omit<BuildExecuteScriptTx, 'signerAddress'>
+
+export interface RunScriptResult extends ExecutionResult {
+}
+
+class Deployments {
+  deployContractResults: Map<string, DeployContractResult>
+  runScriptResults: Map<string, RunScriptResult>
```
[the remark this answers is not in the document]
Lbqds: Hardhat only re-deploys the contract only when the code and constructor params change, so if the code and the constructor params are the same, then the contract will only be deployed once.

Truffle doesn't check this, if you want to deploy the contract twice, you need to write two scripts.

We are more similar to hardhat, we will check the contract code, initial fields, alph amount, and tokens, if any of them changes, we will deploy a new one. I think in most case, if users want to deploy a contract twice, the initial fields should not be the same.

alephium/lib/deployment.ts
```diff
@@ -0,0 +1,389 @@
...
+  deployContract(contract: Contract, params: DeployContractParams): Promise<DeployContractResult>
+  runScript(script: Script, params: RunScriptParams, taskName?: string): Promise<RunScriptResult>
+
+  getDeployContractResult(typeId: string): DeployContractResult
+  getRunScriptResult(typeId: string): RunScriptResult
+}
+
+export type DeployFunction = (deployer: Deployer, networkType: NetworkType) => Promise<void>
+
+export class PrivateKeySigner extends SignerWithNodeProvider {
```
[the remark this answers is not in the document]
Lbqds: Yes, I copied the code from `web3-wallet` because it is private in the previous SDK version, I'll update the SDK to the latest version.

alephium/lib/deployment.ts
```diff
@@ -0,0 +1,389 @@
...
+    const bytecode = script.buildByteCodeToDeploy(params.initialFields ? params.initialFields : {})
+    const codeHash = cryptojs.SHA256(bytecode).toString()
+    const key = taskName ? taskName : script.typeId
+    const previous = runScriptResults.get(key)
+    const tokens = params.tokens ? getTokenRecord(params.tokens) : undefined
+    const needToRun = await needToRunScript(signer.provider, previous, params.attoAlphAmount?.toString(), tokens, codeHash)
+    if (needToRun) {
+      const result = await script.transactionForDeployment(signer, params)
+      await signer.submitTransaction(result.unsignedTx, result.txId)
+      const confirmed = await waitTxConfirmed(signer.provider, result.txId)
```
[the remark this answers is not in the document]
Lbqds: Both hardhat and truffle don't check the confirmation, I guess we can add a config to support this.

alephium/lib/deployment.ts
```diff
@@ -0,0 +1,389 @@
...
+  deployContractResults: Map<string, DeployContractResult>,
+  runScriptResults: Map<string, RunScriptResult>
+): Deployer {
+  const signer = PrivateKeySigner.from(network.mnemonic)
+  const account = signer.getAccountSync()
+
+  const deployContract = async (contract: Contract, params: DeployContractParams): Promise<DeployContractResult> => {
+    // TODO: improve this after migrating to SDK
+    const bytecode = contract.buildByteCodeToDeploy(params.initialFields ? params.initialFields : {})
+    const codeHash = cryptojs.SHA256(bytecode).toString()
```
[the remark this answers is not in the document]
Lbqds: It's different with `contract.codeHash`, it is the hash of `contract.codeHash + initialFields`.

alephium/lib/deployment.ts
Lbqds: Thanks!

alephium/contracts/devnet/deploy_governance.ral
Lbqds: Because the contract ids will change every time we update the contract code, to avoid updating lots of configs, I introduce the devnet deployer, so that the contract ids will be the same except for the `DevnetDeployer` changes, I guess it's time to remove it because the contract code is close to finalized.

alephium/lib/deployment.ts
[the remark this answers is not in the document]
polarker: I see. It's like the `initialStateHash` of Full node. We should add function `Contract.initialStateHash(initialFields)` to calculate the same hash as Full node.

This is good to me for now.

alephium/lib/deployment.ts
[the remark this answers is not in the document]
polarker: Yes, it makes great sense!

Lbqds:
Seems like the latest SDK version(0.2.0-rc.11) doesn't work on nodejs v14.x, and I tested on my local side, maybe we need to remove the nodejs 14 from the CI.

alephium/lib/deployment.ts
polarker: How about this approach:

For each script deployment, we add an optional tag to it and use `Script.typeId -> tag` as the task key. For your example, we will have `RegisterChain -> ETH`, `RegisterChain -> BSC` etc as key.

alephium/lib/deployment.ts
```diff
@@ -0,0 +1,365 @@
...
+      fromGroup: result.fromGroup,
+      toGroup: result.toGroup,
+      txId: result.txId,
+      blockHash: confirmed.blockHash,
+      codeHash: codeHash,
+      attoAlphAmount: params.attoAlphAmount?.toString(),
+      tokens: tokens
+    }
+    runScriptResults.set(key, runScriptResult)
+    return runScriptResult
```
polarker, line 256: Let's log something like the following. Hardhat should have better log messages.
a) Executing script "name"...
b) The execution of script "name" is skipped as it has been executed.

Same for contracts.

alephium/lib/deployment.ts
Lbqds: Sounds great 👍

alephium/lib/deployment.ts
Lbqds: Yes, hardhat uses the [debug](https://github.com/debug-js/debug) library, which can be controlled by the environment variables.

Lbqds:
Now for the idempotency check, we are exactly the same as hardhat:

1. By default we will check the contract code, initial fields, alph amount, tokens, and issue token amount before deploying the contract, and only deploy if these params changed.
2. The user can decide whether to execute the script according to the return value and id.
3. Add an optional `skip`function.

And the hardhat code is at [here](https://github.com/wighawag/hardhat-deploy/blob/32a80276a9d473e43e38b3b07c300e35967b07e8/src/DeploymentsManager.ts#L1190).

alephium/lib/deployment.ts
```diff
@@ -391,13 +392,18 @@ export async function deploy(
...
         migrations.set(script.func.id, Date.now())
       }
-    } catch (error) {
+    } catch (error: any) {
       await saveDeploymentsToFile(deployContractResults, runScriptResults, migrations, network.deploymentFile)
-      throw new Error(`failed to execute deploy script, filepath: ${script.scriptFilePath}, error: ${error}`)
+
+      const errorMsg = error instanceof Error
+        ? error.message
+        : error.error.detail // SDK can't return error, TODO: fix in SDK
```
polarker: [Note]: SDK published

polarker (approved):
LGTM. Great work 👍 I added a few minor updates in another PR.

h0ngcha0 (approved):
LGTM, awesome work! :100:

alephium/lib/deployment.ts
```diff
@@ -0,0 +1,437 @@
...
+}
+
+export type NetworkType = 'mainnet' | 'testnet' | 'devnet'
+
+export interface Configuration<Settings = unknown> {
+  sourcePath?: string
+  artifactPath?: string
+
+  deployScriptsPath: string
+  compilerOptions: CompilerOptions
```
polarker, line 35: Should we make both optional too? The default value for them work in most cases.

alephium/lib/deployment.ts
```diff
@@ -0,0 +1,437 @@
...
+import fs, { promises as fsPromises } from 'fs'
+import * as cryptojs from 'crypto-js'
+
+export interface Network<Settings = unknown> {
+  networkId: number
+  nodeUrl: string
+  mnemonic: string
+  deploymentFile: string
+  confirmations: number
+  settings: Settings
```
polarker, line 25: I am thinking that `networkId`, `deploymentFile`, `confirmations` could also be optional.

Lbqds:
> @Lbqds I added two more comments for better DevX, but we could change it in `alephium-web3`

Got it, I'll update it in the SDK.

PR #28, opened 2022-09-06
Wallet Extension Integration

h0ngcha0:
- Picked Muchen's work for extension wallet integration
- Fixed some UI issues by updating the extension wallet

PR #29, opened 2022-09-07
Fix buf build and update golang to 1.18

Lbqds:
1. Fix failed CI in this [PR](https://github.com/alephium/wormhole-fork/pull/28), we need to update the `buf` to the latest version, which needs Golang v1.18
2. Update the libp2p to the latest version(wormhole did this also in this [PR](https://github.com/wormhole-foundation/wormhole/pull/1518)), and update Golang to 1.18, then we don't need to use different versions of Golang


h0ngcha0 (approved):
LGTM :+1:

polarker (approved):
LGTM

PR #30, opened 2022-09-14
Refund to parent contract and support update wrapped token

Lbqds:
1. Transfer assets to the parent contract during `destroySelf`
2. Support update wrapped token(forgot to add this before)


polarker (commented):
Do we need an governance action to update the refund address? Attackers might get the private key of the address.

Lbqds:
> Do we need an governance action to update the refund address? Attackers might get the private key of the address.

Sounds great 👍 , will update.

alephium/contracts/token_bridge/token_bridge_governance.ral
```diff
@@ -115,4 +116,23 @@ Abstract Contract TokenBridgeGovernance(
...
+        assert!(byteVecSlice!(addressBytes, 0, 1) == #00, ErrorCodes.InvalidUpdateRefundAddressMessage)
+        let newRefundAddress = byteVecToAddress!(addressBytes)
+        let chainSize = u256From1Byte!(byteVecSlice!(payload, 66, 67))
+        assert!(size!(payload) == 67 + chainSize * 2, ErrorCodes.InvalidMessageSize)
+
+        let mut offset = 67
+        for (let mut i = 0; i < chainSize; i = i + 1) {
+            let chainIdBytes = byteVecSlice!(payload, offset, offset + 2)
+            let tokenBridgeForChainId = subContractId!(Path.TokenBridgeForChain ++ chainIdBytes)
+            TokenBridgeForChain(tokenBridgeForChainId).updateRefundAddress(newRefundAddress)
```
polarker, line 134: Is it better to put `refundAddress` in `TokenBridge.ral` so that we only need to update it once?

In most of the cases, token bridge is loaded, so it's not very costly to get refundAddress from `TokenBridge`.

Lbqds: Yes, it's better to put `refundAddress` in `TokenBridge` contract, thanks!

alephium/contracts/token_bridge/token_bridge_governance.ral
```diff
@@ -115,4 +116,20 @@ Abstract Contract TokenBridgeGovernance(
...
+
+    @using(readonly = true, externalCallCheck = false)
+    pub fn getRefundAddress() -> Address {
+        return refundAddress
+    }
+
+    pub fn updateRefundAddress(vaa: ByteVec) -> () {
+        let (targetChainId, payload) = parseAndVerifyGovernanceVAA(vaa, ActionId.UpdateRefundAddress)
+        assert!(targetChainId == localChainId, ErrorCodes.InvalidTargetChainId)
+        let addressBytes = byteVecSlice!(payload, 33, 66)
```
polarker, line 128: Let's maybe support arbitrary address (addressBytesLength ++ addressBytes). I will probably use a multi-sig address for this.

polarker (approved):
LGTM, one last comment

alephium/contracts/token_bridge/token_bridge_governance.ral
Lbqds: Done!

h0ngcha0 (approved):
Looks great! :100:

PR #31, opened 2022-09-14
Minor updates

polarker:
In this PR, I try to put all of the configurations into one file.
1. @Lbqds please help me move other configurations into the same file if I missed any
2. @Lbqds please help me test the changes
3. We may move `configurations.ts` into folder `scripts`. WDYT?


Lbqds:
Thanks for the improvements!

> We may move configurations.ts into folder scripts. WDYT?

In hardhat/truffle, the config file is in the project root path, my thought is that after we move this code to the SDK, we can create the project in a similar way to hardhat/truffle. WDYT?

PR #32, opened 2022-09-15
Add metrices for alephium watcher

Lbqds:


node/pkg/alephium/watcher.go
```diff
@@ -211,6 +248,7 @@ func (w *Watcher) handleConfirmedEvents(logger *zap.Logger, confirmed []*Confirm
 		return nil
 	}
 
+	alphMessagesConfirmed.Add(float64(len(confirmed)))
```
h0ngcha0, line 251: Question: do we wanna do the same in `handleEvents` ?
Or maybe a general question, what is the relationship between `handleObsvRequest` and `fetchEvents/handleEvents` events?

h0ngcha0 (commented):
LGTM :+1: 
Just a question for my understanding.

node/pkg/alephium/watcher.go
Lbqds: There are 4 goroutines in the watcher:

1. `fetchHeight`: used to get the current block height, we can check if the event is confirmed by block height
2. `fetchEvents`: used to fetch events from the full node, and send the unconfirmed events to the `handleEvents` goroutine
3. `handleEvents`: wait for event confirmation, and remove events from the orphan blocks, then confirmed events will be handled in `handleConfirmedEvents` function, so we only need to update the `alphMessagesConfirmed` metrics once
4. `handleObsvRequest`: fetch the event from the full node and re-execute the consensus. It is only used when the event on the chain is not monitored due to unexpected circumstances.

h0ngcha0: Thanks for the explaination! <3

h0ngcha0 (approved):
LGTM :+1:

h0ngcha0:
ALPH metrics seem to be working nicely
https://grafana.softfork.se/d/-GcztLMnz/wormhole-v2?orgId=1

polarker (approved):
LGTM!

PR #34, opened 2022-09-22
Bridge integration tests

Lbqds:


node/cmd/guardiand/node.go
```diff
@@ -163,74 +166,77 @@ func init() {
...
 	alphApiKey = NodeCmd.Flags().String("alphApiKey", "", "Alphium RPC api key")
 	alphContractIds = NodeCmd.Flags().StringSlice("alphContractIds", []string{}, "Alephium contract ids (required)")
 	alphGroupIndex = NodeCmd.Flags().Uint8("alphGroupIndex", 0, "The group index where contracts are deployed (required)")
 	alphMinConfirmations = NodeCmd.Flags().Uint8("alphMinConfirmations", 10, "The min confirmations for alephium tx")
+	alphFetchPeriod = NodeCmd.Flags().Uint8("alphFetchPeriod", 10, "The fetch events period for alwphium watcher")
 
 	logLevel = NodeCmd.Flags().String("logLevel", "info", "Logging level (debug, info, warn, error, dpanic, panic, fatal)")
 
+	devnetGuardianIndex = NodeCmd.Flags().Int("devnetGuardianIndex", 0, "Specify devnet guardian index")
+	integrationTest = NodeCmd.Flags().Bool("integrationTest", false, "Launch node for integration testing")
```
h0ngcha0, line 236: Question: What is the purpose of this flag?

h0ngcha0 (approved):
Only minor / nit comments. Tests are really well written, awesome work! :100:

.github/workflows/it.yaml
```diff
@@ -0,0 +1,68 @@
...
+        run: cd tools && npm install && cd .. && make generate-proto-sdk
+      - name: alephium
+        working-directory: ./alephium
+        # TODO: improve this
+        run: |
+          npm install && cp .env.test .env && npx @alephium/cli devnet start && npx @alephium/cli devnet stop
+          echo "alephium.consensus.block-target-time = 10 millis" >> ~/.alephium-dev/user.conf
+          cat ~/.alephium-dev/user.conf
+          ALEPHIUM_HOME=~/.alephium-dev nohup java -jar ~/.alephium-dev/*.jar &
+          sleep 10
```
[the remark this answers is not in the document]
Lbqds: > I have a script to build those images, maybe i can add on top of this PR.

Very nice, please do, thanks!

node/cmd/guardiand/node.go
Lbqds: These two flags are only used for integration tests:

* devnetGuardianIndex:
On devnet, the guardian key is automatically generated by the guardian index, and the node hostname is in the format:  `guardian-$index`, but on github CI, the hostname is random, so I added the `devnetGuardianIndex` flag. It first tries to get the index from the hostname, and if that fails, it gets the index from the `devnetGuardianIndex` flag.

* integrationTest
The guardian code will lock the memory to protect secret keys from being swapped out to disk, the code is [here](https://github.com/alephium/wormhole-fork/blob/6c1e42adb3b6d86f5afde55d7ee6501bdcc73af8/node/cmd/guardiand/node.go#L304), but github CI does not allow lock the memory, so I added the `integrationTest` flag to skip lock memory.

h0ngcha0: Thanks! I was thinking about the `integrationTest` flag, now i got it. Thanks for the explanation :sun_with_face:

node/pkg/alephium/reobserve.go
```diff
@@ -16,8 +16,14 @@ func (w *Watcher) handleObsvRequest(ctx context.Context, logger *zap.Logger, cli
 		case <-ctx.Done():
 			return
 		case req := <-w.obsvReqC:
-			assume(req.ChainId == uint32(vaa.ChainIDAlephium))
-			assume(len(req.TxHash) == 32)
+			if req.ChainId != uint32(vaa.ChainIDAlephium) {
+				logger.Error("invalid chain id, expect alephium", zap.Uint32("chainId", req.ChainId))
```
polarker, line 20: [Nit]: s/invalid/Invalid/

Similar for other places where the first character of error message is lowercase

polarker (approved):
LGTM, great work ❤️

bridge_ui/src/utils/consts.ts
```diff
@@ -655,20 +655,20 @@ export const ALEPHIUM_BRIDGE_ADDRESS =
     ? "000000000000000000000000000000000000000000000"
     : CLUSTER === "testnet"
     ? "000000000000000000000000000000000000000000000"
-    : "uPDDKWZEgtrJZfvd3i1p6xmULjQRS9mAUVeRQ6eVdP6m";
+    : "26v7DJHTsgtGar18M47KnY7f47WzKCm2oNmLsC3CgpGUb";
```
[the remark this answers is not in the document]
Lbqds: Maybe we need to copy the deployment JSON file to the `bridge_ui` directory because sometimes we run the bridge ui via docker, I'll try it later, if it is possible I'll open another PR for this.

integration_test/alph.ts
```diff
@@ -0,0 +1,267 @@
...
+      throw new Error('Alephium transfer: invalid to address, expect 32 bytes')
+    }
+    return binToHex(toAddress)
+  }
+
+  const getTransactionFee = async (txId: string): Promise<bigint> => {
+    const status = await nodeWallet.provider.transactions.getTransactionsStatus({ txId: txId })
+    // the transaction has been confirmed
+    const blockHash = (status as node.Confirmed).blockHash
+    const block = await nodeWallet.provider.blockflow.getBlockflowBlocksBlockHash(blockHash)
```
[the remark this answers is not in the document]
Lbqds: Yes, it would be more convenient if we could get the transaction directly.

node/pkg/alephium/reobserve.go
Lbqds: Do you mean the error message should start with uppercase? Currently, the log of the guardian node code basically starts with lowercase. And it seems that the golang community prefers to start with lowercase for error messages: https://github.com/golang/go/wiki/CodeReviewComments#error-strings, although this rule does not apply to logs.

polarker: Okay, as long as the style is consistent.

bridge_ui/src/utils/consts.ts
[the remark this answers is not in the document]
polarker: good to me. No need to do it for now.

PR #39, opened 2022-10-28
Run Integration Test with docker-compose

h0ngcha0:
- Simplify CI
- Docker images are needed to run infra in the cloud
- Easier to run locally

### DIFF
--- a/.github/workflows/it.yaml
+++ b/.github/workflows/it.yaml
@@ -22,12 +22,16 @@ jobs:
         run: ./build-docker-images.sh
       - name: start docker images
         working-directory: ./integration_test/docker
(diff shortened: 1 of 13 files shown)


.github/workflows/it.yaml
```diff
@@ -4,7 +4,7 @@ on:
   pull_request:
   push:
     branches:
-      - master
+      - build-docker-images
```
h0ngcha0, line 7: Will update when CI passes

Lbqds: Thanks! I will fix the test after this [PR](https://github.com/alephium/wormhole-fork/pull/37) is merged.

Lbqds (approved):
LGTM 👍 , thanks!

PR #40, opened 2022-11-02
Support ALPH in Explorer and Cloud Functions

h0ngcha0:


Lbqds:
Because we moved the target chain from VAA payload to VAA header, we also need to update the parser:
https://github.com/alephium/wormhole-fork/pull/40/files#diff-f9bc3e27c22610d0f372231ab62442221136daec95a5b2e1b04d0af6e7849f38L80 and https://github.com/alephium/wormhole-fork/pull/40/files#diff-58eede1adacfe9b792ea5333471fcbeb7ee5d1cc87b1ce5c8f3664a91f87e7ecL136

And there are some duplicated codes for the parser, I created an [issue](https://github.com/alephium/wormhole-fork/issues/42) for this. I'll fix it later.

Dockerfile.const
```diff
@@ -34,6 +33,14 @@ COPY scripts ./scripts
 COPY ethereum/.env.test ./ethereum/.env.test
 COPY alephium/.env.test ./alephium/.env.test
 
+ENV NUM_GUARDIANS=$num_guardians
+# WARNING:
+# Credentials file should only be used when generating VAAs for testnet and mainnet, it
+# should never be checked in!
+COPY README.md sensitive/credentials-file.json* .
```
polarker, line 40: Does this mean that we will put the credential files in the docker image?

polarker:
Maybe it's better to have @Lbqds take over this PR and improve the details related to VAA formats. @h0ngcha0 WDYT?

Dockerfile.const
h0ngcha0: It was used to output the env file locally, which is later used to build docker images. But as we discussed the other day, my understanding was off and this is not needed any more.

h0ngcha0:
> Maybe it's better to have @Lbqds take over this PR and improve the details related to VAA formats. @h0ngcha0 WDYT?

Thanks for the help! <3

Lbqds:
Now the explorer and cloud functions work fine, but the explorer still has a lot to improve, I'll improve it later.

integration_test/docker/docker-compose.yaml
```diff
@@ -101,7 +101,7 @@ services:
       - --logLevel=debug
 
   guardian-1:
-    image: alephium/guardiand:0.2.56
+    image: wormhole/guardiand:0.2.56
```
h0ngcha0, line 104: We probably can't push these images to dockerhub, perhaps we should still use the alephium namespace?

Lbqds: Fixed, thanks!

explorer/src/components/ExplorerStats/ExplorerStats.tsx
```diff
@@ -536,81 +521,17 @@ const ExplorerStats: React.FC<StatsProps> = ({
...
-                    ),
-                    imgStyle: { height: 110 },
-                  },
-                  {
-                    header: ChainID[7],
-                    src: oasisIcon,
-                    to: `${explorer}?emitterChain=7`,
+                    header: ChainID[13],
+                    src: alephiumIcon,
+                    to: `${explorer}?emitterChain=255`,
```
polarker, line 526: It seems this is lost: `imgStyle: { height: 110 }`

polarker (approved):
LGTM

alephium/scripts/auto_mine.ts
```diff
@@ -21,7 +21,7 @@ async function mine(num: number): Promise<void> {
   }
   const chainInfo = await nodeProvider.blockflow.getBlockflowChainInfo(chain00)
   console.log(`chain index: ${chain00.fromGroup} -> ${chain00.toGroup}, height: ${chainInfo.currentHeight}`)
-  await new Promise((resolve) => setTimeout(resolve, 1000))
+  await new Promise((resolve) => setTimeout(resolve, 10000))
```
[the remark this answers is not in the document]
h0ngcha0: This is updated since in the cloud 1s block makes the full node unresponsive to the new transactions after a while (~30min~1hr)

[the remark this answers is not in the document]
polarker: Let's roll this back to 1000 and use full node v1.5.2. It should work, otherwise we will investigate.

explorer/src/components/ExplorerStats/ExplorerStats.tsx
h0ngcha0: it is not removed, as we can see [here](https://github.com/alephium/wormhole-fork/pull/40/files/6d6121f084c8585f3b9271159602f286a83f6aaf#diff-1243e033d2e07c5bbbfd4d02c9599bba11f348d05d7d9e37b1169a5941aa15bdR527-R537)

Lbqds:
I added two commits to fix the cloud functions because we support transfer tokens to multi-sig addresses.

h0ngcha0:
> I added two commits to fix the cloud functions because we support transfer tokens to multi-sig addresses.

Thanks!

PR #43, opened 2022-11-09
Rename guardian module path

Lbqds:
Rename the module path to make it available as a dependency library, and it is necessary for #40


node/pkg/proto/node/v1/node.pb.go
```diff
@@ -0,0 +1,1810 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
```
h0ngcha0, line 2: Do we need to check in all the auto-generated go code?

h0ngcha0 (commented):
LGTM, except for one question.

node/pkg/proto/node/v1/node.pb.go
Lbqds: Yes, if other go projects need it, then we don't need to regenerate the code, wormhole also does this: https://github.com/wormhole-foundation/wormhole/tree/dev.v2/node/pkg/proto

h0ngcha0 (approved):

polarker (approved):
LGTM

PR #49, opened 2022-11-17
Upgrade to full node 1.5.2

Lbqds:


alephium/package.json
```diff
@@ -10,9 +10,9 @@
     "lint": "eslint . --ext ts",
     "lint:fix": "eslint . --fix --ext ts",
     "build": "rm -rf dist && npx tsc --build .",
-    "compile-contracts": "npx @alephium/cli@latest compile",
-    "test": "npx @alephium/cli@latest test -i -v",
-    "deploy:devnet": "npx @alephium/cli@latest deploy -n devnet",
+    "compile-contracts": "npx @alephium/cli compile",
+    "test": "npx @alephium/cli test -i -v",
+    "deploy:devnet": "npx @alephium/cli deploy -n devnet",
```
Lbqds, line 15: Both `@alephium/cli@latest` and `@alephium/cli` work in my local environment, but `@alephium/cli@latest` does not work on GitHub CI, I don't know why.

polarker: Github action probably uses its own package manager to save network bandwidth. We could use hardcoded version instead of `@latest`.

polarker (approved):
LGTM

PR #56, opened 2023-04-24
Integrate the new explorer

Lbqds:
The new explorer backend uses mongodb to replace bigtable.


explorer-backend/fly/utils/coin_gecko.go
```diff
@@ -73,3 +74,69 @@ func FetchCoinGeckoCoin(chainId vaa.ChainID, symbol, name string) *CoinGeckoCoin
...
+		}
+		return nil, parseErr
+	}
+	if len(parsed.Prices) >= 1 {
+		var priceIndex int
+		if withinLastHour {
+			// use the last price in the list, latest price
+			priceIndex = len(parsed.Prices) - 1
+		} else {
+			// use a price from the middle of the list, as that should be
```
h0ngcha0, line 126: Q: Could it be inaccurate if timestamp is 2 hours ago and we are taking price from 7 hours ago? (IIUC)

explorer-backend/fly/storage/repository.go
```diff
@@ -253,8 +254,8 @@ func (s *Repository) upsertTokenTransfer(ctx context.Context, v *vaa.VAA, tokenT
 	return err
 }
 
-func (s *Repository) getToken(ctx context.Context, tokenAddress string) (*TokenUpdate, error) {
-	filter := bson.D{{Key: "tokenAddress", Value: tokenAddress}}
+func (s *Repository) getToken(ctx context.Context, tokenAddress string, tokenChain vaa.ChainID) (*TokenUpdate, error) {
+	filter := bson.D{{Key: "_id", Value: getTokenId(tokenAddress, tokenChain)}}
```
h0ngcha0, line 281: Q: `_id` seems to be the built-in id for Mongo with timestamps, a bit of randomness and an incremental counter. Do we want to replace that with our own id or could it be better to create a `tokenId` field?

h0ngcha0: Did some research, seems to be fine to have custom value in our case since it is always deterministic and unique.

explorer-backend/fly/migration/migration.go
```diff
@@ -164,5 +170,44 @@ func Run(db *mongo.Database) error {
...
+			},
+		},
+		{
+			Keys: bson.D{
+				{Key: "date", Value: 1},
+				{Key: "emitterChain", Value: 1},
+				{Key: "emitterAddr", Value: 1},
+			},
+		},
+		{
```
h0ngcha0, line 188: Same as a previous comment, seems that we only need to create the first and the last compound indexes

h0ngcha0 (commented):
Great work, some questions and comments :1st_place_medal:

explorer-backend/api/handlers/guardian/guardian_set.go
```diff
@@ -0,0 +1,87 @@
...
+	Index: 2,
+	Keys: []eth_common.Address{
+		eth_common.HexToAddress("0x58CC3AE5C097b213cE3c81979e1B9f9570746AA5"), // Certus One
+		eth_common.HexToAddress("0xfF6CB952589BDE862c25Ef4392132fb9D4A42157"), // Staked
+		eth_common.HexToAddress("0x114De8460193bdf3A2fCf81f86a09765F4762fD1"), // Figment
+		eth_common.HexToAddress("0x107A0086b32d7A0977926A205131d8731D39cbEB"), // ChainodeTech
+		eth_common.HexToAddress("0x8C82B2fd82FaeD2711d59AF0F2499D16e726f6b2"), // Inotel
+		eth_common.HexToAddress("0x11b39756C042441BE6D8650b69b54EbE715E2343"), // HashQuark
+		eth_common.HexToAddress("0x54Ce5B4D348fb74B958e8966e2ec3dBd4958a7cd"), // ChainLayer
+		eth_common.HexToAddress("0x66B9590e1c41e0B226937bf9217D1d67Fd4E91F5"), // FTX
```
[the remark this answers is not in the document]
Lbqds: I removed the hardcoded `guardianSets`, and now it will fetch `guardianSets` from the guardian node.

explorer-backend/fly/guardiansets/gst_data.go
```diff
@@ -67,80 +73,66 @@ func (gs *GuardianSets) UpdateGuardianSet(ctx context.Context) {
...
 		return nil
 	}
-	guardianSetAddresses := make([]eth_common.Address, len(addresses))
-	for i, address := range addresses {
-		guardianSetAddresses[i] = eth_common.HexToAddress(address)
+
+	gs.lock.Lock()
+	defer gs.lock.Unlock()
+
+	maxGuardianSetIndex := guardianSets[len(guardianSets)-1].Index
```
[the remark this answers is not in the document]
Lbqds: Yes, it is guaranteed by two parts:

1. The contract guarantees that the index of the updated guardian set is increasing.
2. When the explorer-backend starts up, it queries all historical guardian sets on the chain.

explorer-backend/fly/migration/migration.go
```diff
@@ -136,5 +141,28 @@ func Run(db *mongo.Database) error {
 		return err
 	}
 
+	tokenTransferIndexes := []mongo.IndexModel{
+		{
+			Keys: bson.D{{Key: "tokenAddress", Value: 1}},
+		},
+		{
```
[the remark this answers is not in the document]
Lbqds: Thanks for the information! But the doc says that in this case, it won't be as efficient as standalone indexing:

```
MongoDB can also use the index to support a query on the item and stock fields, since the item field corresponds to a prefix. However, in this case the index would not be as efficient in supporting the query as it would be if the index were on only item and stock.
```

explorer-backend/fly/utils/coin_gecko.go
Lbqds: This code snippet is from the old explorer, and I had the same question as you. Although the prices returned in my testing are mainly distributed by the hour, the Coingecko documentation does not specify the time interval at which prices will be returned. I'll do more tests later.

explorer-backend/fly/storage/repository.go
Lbqds: Yes, in this case, the `_id` is the `tokenAddress:tokenChain`, which is unique.

explorer/.env.development
```diff
@@ -1,8 +1,8 @@
-GATSBY_SITE_URL=http://localhost:8000
-GATSBY_GA_TAG=G-tag-goes-here
+GATSBY_SITE_URL=http://localhost:8001
+GATSBY_GA_TAG=not_available
 
-GATSBY_BIGTABLE_FUNCTIONS_DEVNET_BASE_URL=https://europe-west1-alephium-wormhole.cloudfunctions.net/alephium_wormhole_events_database/
-GATSBY_BIGTABLE_FUNCTIONS_TESTNET_BASE_URL=http://not_available
-GATSBY_BIGTABLE_FUNCTIONS_MAINNET_BASE_URL=http://not_available
+GATSBY_DEVNET_BACKEND_URL=http://192.168.2.7:8100/
```
polarker, line 4: Why do we have this constant?

explorer/src/utils/explorer.ts
```diff
@@ -135,6 +138,9 @@ const nativeExplorerContractUri = (
       base = "https://ftmscan.com/address/";
     } else if (chainId === chainIDs["aurora"]) {
       base = "https://aurorascan.dev/address/";
+    } else if (chainId === chainIDs["alephium"]) {
+      // FIXME
+      base = "https://explorer.testnet.alephium.org/address"
```
polarker, line 143: Is this fixed?

explorer-backend/fly/guardiansets/gst_data.go
```diff
@@ -51,9 +45,21 @@ func NewGuardianSets(
 	}
 }
 
-func (gs *GuardianSets) GetGuardianSet(index int) (*common.GuardianSet, error) {
+func (gs *GuardianSets) GetGuardianSet(ctx context.Context, index int) (*common.GuardianSet, error) {
```
polarker, line 50: It seems WH uses hardcoded guardian sets for mainnet. We may need to do the same later?

explorer-backend/fly/utils/coin_gecko.go
```diff
@@ -86,52 +76,126 @@ func rangeFromTime(t time.Time, hours int) (start time.Time, end time.Time) {
...
 	}
 	return 0, fmt.Errorf("no price found for %v", coinId)
 }
+
+type Price struct {
+	USD float64 `json:"usd"`
+}
+type CoinGeckoCoinPrices map[string]Price
+
+func fetchCoinGeckoPrices(coinIds []string) (map[string]float64, error) {
```
polarker, line 142: Why does the backend needs to fetch token price? It seems to be the work of front-end

node/cmd/guardiand/adminserver.go
```diff
@@ -559,12 +559,11 @@ func (s *nodePrivilegedService) SendObservationRequest(ctx context.Context, req
 }
 
 func (s *nodePrivilegedService) GetNextGovernanceVAASequence(ctx context.Context, req *nodev1.GetNextGovernanceVAASequenceRequest) (*nodev1.GetNextGovernanceVAASequenceResponse, error) {
-	maxSequence, err := s.db.MaxGovernanceVAASequence(s.governanceChainId, s.governanceEmitterAddress)
+	nextSequence, err := s.db.NextGovernanceVAASequence(s.governanceChainId, s.governanceEmitterAddress)
```
polarker, line 562: 1. It seems WH does not have such data in the database?
2. If different guardians have different db due to bug or network issues, will they produce different next sequence here?

proto/publicrpc/v1/publicrpc.proto
```diff
@@ -77,6 +83,41 @@ message GetSignedVAAResponse {
...
+message GetGovernanceVAABatchResponse {
+  message Entry {
+    ChainID target_chain = 1;
+    uint64 sequence = 2;
+    bytes vaa_bytes = 3;
+  }
+
+  repeated Entry entries = 1;
+}
+
```
polarker, line 120: Why do we need these new endpoints (not from WH)?

explorer/.env.development
Lbqds: It should be `localhost`, will fix it.

explorer/src/utils/explorer.ts
Lbqds: Yes, it has been fixed here: https://github.com/alephium/wormhole-fork/pull/59/files#diff-8b4475005d0f7c6a2c534389d362314bae61a91d4a2d04a9b679da46c082dd42R151

explorer-backend/fly/guardiansets/gst_data.go
Lbqds: Yes, we can also hardcode the guardian sets for mainnet.

explorer-backend/fly/utils/coin_gecko.go
Lbqds: The code is migrated from the old wormhole explorer-backend. The purpose of fetching the token price is for statistics. When the explorer-backend receives a new VAA, it will get the token price and then update the statistics.

node/cmd/guardiand/adminserver.go
Lbqds: >It seems WH does not have such data in the database?

Wormhole also stores governance VAA data, but wormhole does not provide a function similar to `GetNextGovernanceVAASequence`. For governance VAA, the difference between us and Wormhole is that our `Governance` contract and `TokenBridge` contract share the same sequence. I added this function to facilitate getting the next sequence vaa on the guardian.

>If different guardians have different db due to bug or network issues, will they produce different next sequence here?

It will produce a different next sequence if the guardian has a different db. Therefore, the most reliable way to get the next governance sequence is to query the contract state. Maybe we should remove it from the guardian and add a command line to the `clients/js`, WDYT?

proto/publicrpc/v1/publicrpc.proto
Lbqds: This is used by the explorer-backend to sync missing VAAs from the guardian. The wormhole explorer-backend currently does not support syncing missing VAAs.

node/cmd/guardiand/adminserver.go
polarker: That's a good approach I think as it reduces Guardian's code

polarker (approved):
LGTM

PR #57, opened 2023-05-03
Remove memory locking

h0ngcha0:
Reason for removing memory lock:

Our signing key is done through KMS, they will never end up in memory.

### DIFF
--- a/integration_test/docker/docker-compose.yaml
+++ b/integration_test/docker/docker-compose.yaml
@@ -77,7 +77,6 @@ services:
       - "1"
       - --devnetGuardianIndex
       - "0"
-      - --integrationTest
       - --publicRPC
       - "[::]:7070"
       - --publicWeb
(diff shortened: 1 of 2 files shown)


Lbqds (approved):
LGTM, thanks!

h0ngcha0:
> @h0ngcha0 Is memory locking blocking anything? Otherwise good to keep I think.

I was looking at an `IPC_LOCK` issue that makes the memory locking fail, which was not the case when we run the Guardian node before. During the investigation I realized that the reason that memory lock was required was because of the signing key, so actually we do not need this.

I will fix the `IPC_LOCK` issue if we want to keep this.

h0ngcha0:
> @h0ngcha0 Is memory locking blocking anything? Otherwise good to keep I think.

This is fixed now, was caused by [a small but tricky issue](https://github.com/alephium/wormhole-fork/pull/59/commits/3337748cb04fd6818203a1843ff3007a7c6d4e20) from Dockerfile.

Closing this PR.

PR #59, opened 2023-05-05
Deploy to Testnet

h0ngcha0:
Some updates while deploying to testnet


relayer/spy_relayer/.env.tilt
```diff
@@ -14,11 +14,17 @@ RELAYER_CHAINS='[{
...
+  "chainName": "BSC",
+  "nativeCurrencySymbol": "BNB",
+  "nodeUrl": "http://bsc-devnet:8545",
+  "walletPrivateKeys": ["0x4f3edf983ac636a65a842ce7c78d9aa706d3b113bce9c46f30d7d21715b23b1d"]
 }, {
   "chainId": 255,
   "nativeCurrencySymbol": "ALPH",
   "nodeUrl": "http://alph-full-node:22973",
-  "walletPrivateKeys": ["vault alarm sad mass witness property virus style good flower rice alpha viable evidence run glare pretty scout evil judge enroll refuse another lava"]
+  "walletPrivateKeys": ["a642942e67258589cd2b1822c631506632db5a12aabcf413604e785300d762a5"]
```
polarker, line 27: Private keys are only for devnet?

h0ngcha0: Yeah, devnet only

h0ngcha0:
> General question: if we want to deploy to mainnet, how much changes would it be?

The change should not be that big. We need to deploy all contracts to mainnet and update the corresponding configs. Infra should be almost the same between testnet and mainnet. Probably there will be some work to get our partners up and running for Guardians as well.

polarker (approved):
LGTM, ready fro merge

PR #60, opened 2023-05-08
Update ETH watcher

Lbqds:


h0ngcha0 (commented):
LGTM, will test it in testnet before approval.

PR #65, opened 2023-06-09
Fix never ending loading error

Lbqds:
This PR fixed the issue reported [here](https://discord.com/channels/747741246667227157/948144672402972682/1116406209340256337), we need to properly handle the errors that occur when fetching token information from the source chain.


polarker (approved):
LGTM

PR #66, opened 2023-06-12
Bridge reward

Lbqds:


alephium/contracts/token_bridge/bridge_reward_router.ral
```diff
@@ -0,0 +1,20 @@
+Contract BridgeRewardRouter() {
+    @using(preapprovedAssets = true, assetsInContract = true, checkExternalCaller = false)
+    pub fn completeTransfer(tokenBridgeForChain: TokenBridgeForChain, vaa: ByteVec, caller: Address) -> () {
+        tokenBridgeForChain.completeTransfer{caller -> ALPH: dustAmount!()}(vaa, caller)
```
Lbqds, line 4: If the sequence of vaa is too large, the `completeTransfer` function will return directly. In this case, we will still issue a reward to the user. But the cost for an attacker to exploit this method would be high. The attacker need to send 768 consecutive transactions on the source chain to receive a reward of 1 ALPH. So maybe it's not a problem.

However, if we still want to avoid this issue, we may need to change the `TokenBridgeForChain` contract, such as changing the `completeTransfer` function to return a value indicating if the redemption is successful. But this requires upgrading the contract on the testnet. I'm not sure if there is a way to determine the success of redemption without changing the `TokenBridgeForChain` contract.

alephium/contracts/token_bridge_scripts/complete_transfer.ral
```diff
@@ -1,7 +1,7 @@
 TxScript CompleteTransfer(
+    bridgeRewardRouter: BridgeRewardRouter,
     tokenBridgeForChain: TokenBridgeForChain,
     vaa: ByteVec
 ) {
-    let caller = callerAddress!()
-    tokenBridgeForChain.completeTransfer{caller -> ALPH: dustAmount!()}(vaa, caller)
+    bridgeRewardRouter.completeTransfer{callerAddress!() -> ALPH: dustAmount!()}(tokenBridgeForChain, vaa, callerAddress!())
```
polarker, line 6: We should have two versions: the old `CompleteTransfer` and a new `CompleteTransferWithReward`. We will need to switch back to `CompleteTransfer` after the bootstrap phase.

alephium/contracts/token_bridge_scripts/complete_transfer_router.ral
```diff
@@ -1,7 +0,0 @@
-TxScript CompleteTransfer(
```
polarker: Let's rename this as `CompleteTransferWithReward`

alephium/contracts/token_bridge/bridge_reward_router.ral
polarker: We decided not to work on this.

alephium/artifacts/ts/scripts.ts
```diff
@@ -69,10 +70,14 @@ export const AttestToken = new ExecutableScript<{
   consistencyLevel: bigint;
 }>(Script.fromJson(AttestTokenScriptJson));
 export const CompleteTransfer = new ExecutableScript<{
-  bridgeRewardRouter: HexString;
   tokenBridgeForChain: HexString;
   vaa: HexString;
 }>(Script.fromJson(CompleteTransferScriptJson));
+export const CompleteTransferWithReward = new ExecutableScript<{
```
polarker, line 76: This is not used in this commit. So it's not being used right?

polarker: Resolved. It's in the next commit.

polarker (approved):
LGTM

h0ngcha0 (approved):
LGTM :+1:

PR #67, opened 2023-06-15
Fix build explorer image

Lbqds:


explorer-backend/transactions/evm.go
```diff
@@ -99,10 +99,50 @@ func (w *EVMWatcher) fetchEventsFromBlockRange(ctx context.Context, fromHeight u
...
 
-	for height := fromHeight; height < toHeight; height++ {
-		if err := w.fetchEventsByBlockNumber(ctx, height); err != nil {
+	client := w.connector.Client()
+	query := eth.FilterQuery{
+		FromBlock: big.NewInt(int64(fromHeight)),
+		ToBlock:   big.NewInt(int64(toHeight)),
+		Addresses: []ethCommon.Address{*w.contractAddress},
+		Topics:    [][]ethCommon.Hash{{ethereum.LogMessagePublishedTopic}},
+	}
```
Lbqds, line 108: Since there are many blocks at the first startup, I just tested it on the testnet, if it is very slow to get one by one, so I changed it to get events in batches.

node/Dockerfile
```diff
@@ -25,7 +25,7 @@ ADD ./node .
 RUN if [ "$network" != "devnet" ]; then apt-get update && apt-get install libcap2-bin -y; fi
 
 RUN --mount=type=cache,target=/root/.cache --mount=type=cache,target=/go \
-  go build -race -gcflags="all=-N -l" --ldflags '-extldflags "-Wl,--allow-multiple-definition"' -mod=readonly -o /guardiand github.com/alephium/wormhole-fork/node
+  go build -gcflags="all=-N -l" --ldflags '-extldflags "-Wl,--allow-multiple-definition"' -mod=readonly -o /guardiand github.com/alephium/wormhole-fork/node
```
Lbqds, line 28: The purpose of disabling the race detector is to solve the OOM(out of memory) error. Currently, the guardian enables the race detector at runtime to detect memory read-write conflicts. And I did some tests on my local side:

1. If we enable the race detector and lock memory, it leads to an OOM error.
2. If we only enable the race detector but without locking memory, the memory usage is relatively stable, around 300-400MB.
3. If we only lock memory but disable the race detector, the memory usage is also relatively stable, but much higher than case 2, around 2.5GB.

I checked the latest wormhole code, I found that they made the `-race` flag [optional](https://github.com/wormhole-foundation/wormhole/blob/21276f7f752b2f538bef49ace3046d7084bfe641/node/Dockerfile#L23) and no longer [lock memory on the testnet](https://github.com/wormhole-foundation/wormhole/blob/21276f7f752b2f538bef49ace3046d7084bfe641/node/cmd/guardiand/node.go#L419).

Additionally, [some issues](https://github.com/golang/go/issues/37233) suggest that the golang race detector may have memory leaks.

And OOM would not happen again after deploying the new code to testnet.

h0ngcha0 (approved):
LGTM, great investigation :+1:

explorer-backend/transactions/evm.go
Lbqds: 1. No, for the EVM chains, we query from the height of contract deployments; for Alephium, we query from the event index 0.
2. In order to avoid losing events, it will query every reboot, but not from the begining. For example, if we deploy the contracts at a block with a height of `A`, then the explorer backend stops running at a height of `B`, and restarts at a height of `C`, where `C > B > A`, then it will start querying events from B instead of start with A

polarker (approved):

explorer-backend/transactions/evm.go
polarker: If the explorer backend stops at a height of `B` with a block `b1` and there are two blocks `b0` and `b1` at the same height with `b0` becomes the cannonical block eventually, does the sync function work well after rebooting?

Lbqds: No, but I think Alephium does not have this problem because we query from the last event index.

For EVM chains, I think we can start the query event from the height `B - N` when restarting, where N can be 64(2 epochs) for ETH. WDYT?

polarker: Yes, that sounds great to me!

explorer-backend/transactions/watcher.go
```diff
@@ -27,6 +27,30 @@ func NewWatcher(logger *zap.Logger, configs *common.BridgeConfig, db *mongo.Data
...
+}
+
+func (w *Watcher) GetLatestEventIndexEvm(ctx context.Context, chainId vaa.ChainID, confirmations uint32) (*uint32, error) {
+	eventIndex, err := w.GetLatestEventIndex(ctx, chainId)
+	if err != nil {
+		return nil, err
+	}
+	fromIndex := uint32(0)
+	if *eventIndex > confirmations {
+		fromIndex = *eventIndex - confirmations
```
polarker, line 49: Do we need `fromIndex = max(1, *eventIndex - confirmations` so it's always positive?

Lbqds: The `fromIndex >=0` is always true, and since we use `uint32` here, if `eventIndex` is less than `confirmations`, then it will underflow, and we will get a big `uint32` number.

If we want `max(1, *eventIndex - confirmations)`, we need to do type conversion: `fromIndex := uint32(maxInt64(1, int64(*eventIndex) - int64(confirmations)))`.

Lbqds: And although golang already supports generics, there is still no elegant way to write a generic comparison function.

polarker: Then we should probably do something like:
```
if (*eventIndex <= confirmations) 1 else (*eventIndex - confirmations)
```

PR #69, opened 2023-06-19
Improve fetch txs status

Lbqds:
Reduce the number of requests when getting transaction status.


bridge_ui/src/hooks/useAlephiumWallet.ts
```diff
@@ -26,7 +26,8 @@ export function useAlephiumWallet() {
 
   useEffect(() => {
     if (context.account !== undefined && context.signerProvider?.nodeProvider !== undefined) {
-      const wallet = new AlephiumWallet(context.signerProvider, context.signerProvider.nodeProvider, context.account.address)
+      const nodeProvider = new NodeProvider(context.signerProvider?.nodeProvider)
```
Lbqds, line 29: This can be removed when we update the extension wallet's `@alephium/web3` dependency to 0.13.0.

polarker: What change did we introduce is relevant to this comment?

polarker (approved):
LGTM

PR #70, opened 2023-06-22
Publish wormhole sdk to @alephium/wormhole-sdk

Lbqds:
We need to set the `NPM_TOKEN` secret. cc @polarker

### DIFF
--- a/.github/workflows/publish-sdk.yml
+++ b/.github/workflows/publish-sdk.yml
@@ -20,15 +20,42 @@ jobs:
         with:
           go-version: '1.19.2'
 
-      - name: Build
+      - name: Get package version from package.json
+        id: getPackageVersion
+        run: echo "::set-output name=version::$(node -p 'require("./sdk/js/package.json").version')"
+
+      - name: Check if version is published
+        id: checkVersion
+        run: |
+          VERSION=${{ steps.getPackageVersion.outputs.version }}
+          set +e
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-sdk.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Build
+        run: |
+          cd tools && npm ci && ./build.sh && cd ..
+          tools/bin/buf generate --template buf.gen.web.yaml
+          cd ethereum && npm ci && cd ..
+          cd sdk/js && npm ci && npm run build
+
+      - name: Publish
+        uses: JS-DevTools/npm-publish@v2
```
polarker, line 31: Let's try to publish with npm directly and don't rely on a third party tool which might not be risky sometimes.

polarker:
`NPM_TOKEN` is enabled now.

.github/workflows/publish-sdk.yml
Lbqds: Done! I chose to use this because we wanted to publish based on the version defined in package.json, rather than using git tags(because we may not update the SDK code when we create a new tag). And `JS-DevTools/npm-publish@v2` can check if the version has been published. Now, we are using npm view to check if the version has been published.

polarker: Agree that it's not convenient to publish when there are tag versions and npm package versions.

PR #74, opened 2023-07-07
Add bridge token list

Lbqds:


token-list/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+      tokenId,
+      deployments.groupIndex
+    )
+  }
+
+  const getTokenMetadata = async (tokenId: string): Promise<TokenMetaData> => {
+    if (tokenId === ALPH_TOKEN_ID) {
+      return { name: ALPH.name, symbol: ALPH.symbol, decimals: ALPH.decimals }
+    }
+    const nodeProvider = web3.getCurrentNodeProvider()
```
h0ngcha0, line 43: Do you think it makes sense to consult [Token List](https://github.com/alephium/token-list) first here?

h0ngcha0 (commented):
LGTM :+1: 

Is the next step to publish it somewhere like [this](https://gnosis.blockscout.com/bridged-tokens/eth)?

Lbqds:
> Is the next step to publish it somewhere like [this](https://gnosis.blockscout.com/bridged-tokens/eth)?

Yeah, wormhole uses a dedicated repo to track all bridge tokens: https://github.com/wormhole-foundation/wormhole-token-list

token-list/utils.ts
Lbqds: Good idea, I have added a check to verify if the token is in the token list, but the token metadata is still queried from the chain.

h0ngcha0 (approved):
LGTM :+1:

polarker (approved):
LGTM

PR #75, opened 2023-08-18
Update @alephium/web3 to 0.18.0

Lbqds:


.github/workflows/it.yaml
```diff
@@ -20,6 +20,8 @@ jobs:
           node-version: '16'
       - name: build devnet docker images
         run: ./build-docker-images.sh "devnet"
+      - name: docker system prune
+        run: docker system prune -f
```
Lbqds, line 24: This is to avoid integration tests failing due to not enough disk space.

PR #77, opened 2023-08-31
Remove the `alphMinConfirmations` flag

Lbqds:
1. Address this comment: https://github.com/alephium/bridge-guardian-infra/pull/5/files#r1309815195
2. Improve the creation of `LocalTokenPool` on Alephium

I will release 0.1.0 after this PR is merged.


node/pkg/alephium/reobserve.go
```diff
@@ -144,7 +144,7 @@ func (w *Watcher) getGovernanceEventsByTxId(
 
 		reobservedEvents = append(reobservedEvents, &reobservedEvent{
 			&event,
-			maxUint8(msg.consistencyLevel, w.minConfirmations),
+			msg.consistencyLevel,
```
polarker, line 147: Not related to this PR: how do we enforce the minimum consistencyLevel for new messages from users?

h0ngcha0 (approved):
LGTM

node/pkg/alephium/reobserve.go
Lbqds: Currently, we enforce the minimum consistency level in both frontend and contract:

1. The frontend sets the `consistencyLevel` to 105 when the user sends a transaction
2. We also check in the contract that the `consistencyLevel` for the transfer transaction is not less than 105

PR #78, opened 2023-09-08
Reduce the number of requests

Lbqds:


h0ngcha0 (commented):
LGTM :+1:

explorer-backend/transactions/evm.go
```diff
@@ -151,47 +151,6 @@ func (w *EVMWatcher) fetchEventsFromBlockRange(ctx context.Context, fromHeight u
...
-	w.logger.Info("new bridge transaction", zap.String("sender", *sender), zap.String("txId", log.TxHash.Hex()))
-	return &BridgeTransaction{
-		vaaId:      vaaId,
-		txId:       log.TxHash.Hex(),
-		address:    *sender,
-		eventIndex: uint32(log.BlockNumber),
-	}, nil
-}
-
-func (w *EVMWatcher) fetchEventsByHeader(ctx context.Context, header *types.Header) error {
```
polarker, line 192: We could pass the client as function argument to make the diff smaller

node/pkg/ethereum/watcher.go
```diff
@@ -169,32 +171,21 @@ func (w *Watcher) Run(ctx context.Context) error {
...
-			return fmt.Errorf("invalid poll interval setting")
-		}
-		w.ethConn, err = NewBlockPollConnector(ctx, baseConnector, time.Duration(*w.pollIntervalMs)*time.Millisecond, true)
-		if err != nil {
-			ethConnectionErrors.WithLabelValues(w.networkName, "dial_error").Inc()
-			p2p.DefaultRegistry.AddErrorCount(w.chainID, 1)
-			return fmt.Errorf("creating block poll connector failed: %w", err)
-		}
-	} else {
-		w.ethConn, err = NewEthereumConnector(timeout, w.networkName, w.url, w.contract, logger)
```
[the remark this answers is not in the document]
Lbqds: Yes, using the block poller in devnet makes testing and debugging more convenient. And it is the same as on testnet/mainnet except for `useFinalizedBlock`.

node/pkg/alephium/watcher.go
```diff
@@ -119,13 +120,22 @@ func NewAlephiumWatcher(
...
+		blockPollerEnabled: &atomic.Bool{},
+		pollIntervalMs:     pollIntervalMs,
 
 		client: NewClient(url, apiKey, 10),
 	}
 	return watcher, nil
 }
 
+func (w *Watcher) DisableBlockPoller() {
+	w.blockPollerEnabled.Store(false)
```
polarker, line 132: If there are 2 poll requests, the block poller will be enabled. Once one of it is end, the flag will be set false. Will the second poll request be terminated too?

Lbqds: No, the poll request is not for each individual transaction. It continues to poll the block height as long as there are pending transactions and only stops polling the block height when there are no pending transactions.

explorer-backend/transactions/evm.go
Lbqds: I have already merged these two functions, so we don't need to pass the client as a parameter. Actually I made very few code changes, I simply moved a few lines of code. The extensive diff here is mainly because the `bridgeTxFromLog` function is located between these two functions, but the `bridgeTxFromLog` function remains unchanged.

polarker: Okay

node/pkg/alephium/watcher.go
polarker: perfect

polarker (approved):
LGTM

PR #80, opened 2023-09-21
Check the vaa confirmations based on timestamp

Lbqds:


node/pkg/alephium/watcher.go
```diff
@@ -366,7 +368,8 @@ func (w *Watcher) handleEvents_(
 	pendingEvents := map[string]*UnconfirmedEventsPerBlock{}
 
 	process := func(height int32) error {
-		logger.Debug("processing events", zap.Int32("height", height))
+		now := time.Now().UnixMilli()
+		logger.Debug("processing events", zap.Int32("height", height), zap.Int64("now", now))
```
h0ngcha0, line 374: [Nit] I guess logger has the current timestamp already

node/pkg/alephium/watcher.go
```diff
@@ -385,16 +388,17 @@ func (w *Watcher) handleEvents_(
...
-				if blockEvents.header.Height+int32(eventConfirmations) > height {
-					logger.Debug(
-						"event not confirmed",
-						zap.String("txId", event.TxId),
-						zap.Int32("blockHeight", blockEvents.header.Height),
-						zap.Uint8("confirmations", eventConfirmations),
-					)
+				consistencyLevel := event.msg.consistencyLevel
+				if !isEventConfirmed(consistencyLevel, blockEvents.header, now, height) {
+					logger.Debug("event not confirmed", zap.String("txId", event.TxId), zap.Uint8("consistencyLevel", consistencyLevel))
```
h0ngcha0, line 403: [Nit] Would it easier to read the log if we log inside the `isEventConfirmed`, so we know if it's not confirmed due to height or timestamp?

integration_test/src/alph.ts
```diff
@@ -82,7 +82,7 @@ export async function createAlephium(): Promise<AlephiumBridgeChain> {
   const bridgeRewardRouterAddress = addressFromContractId(bridgeRewardRouterId)
   const sequence = new Sequence()
   const defaultArbiterFee = 0n
-  const defaultConfirmations = 1
+  const defaultConfirmations = 0
```
h0ngcha0, line 85: Question: why do we need to set the the default confirmation to `0` here?
Could you please remind me on the reasoning behind using timestamp?

h0ngcha0 (commented):
LGTM
Only Nit comments and questions.

node/pkg/alephium/watcher.go
Lbqds: Sounds great 👍, will update.

integration_test/src/alph.ts
h0ngcha0: Thanks a lot for the great explanation :+1:

h0ngcha0 (approved):
LGTM

Lbqds:
Close this PR because it has already been merged, but github didn't automatically close it.

PR #84, opened 2023-10-05
Publish Docker Images

h0ngcha0:
Automatically publish docker images based on tags

CI is already configured using workload identity so it has enough permission to push images to GCR.


.github/workflows/publish-docker-images.yml
```diff
@@ -0,0 +1,47 @@
...
+        fetch-depth: 0
+    - id: 'auth'
+      uses: 'google-github-actions/auth@v0'
+      with:
+        workload_identity_provider: projects/1004240933630/locations/global/workloadIdentityPools/github-pool/providers/github-provider
+        service_account: wormhole-fork-repo@alephium-org.iam.gserviceaccount.com
+    - name: auth docker
+      run: gcloud auth configure-docker
+    - name: build and push testnet docker images
+      run: ./build-docker-images.sh "testnet" ${{ needs.get_dist_tag.outputs.DIST_TAG }} true
```
polarker, line 47: We will need to change to this mainnet later?

polarker:
Another important question: each tag now will trigger the publish of both docker image and SDK right?

Since the docker image would be updated less often or differently from the SDK. Maybe we use a prefix for docker image publish? For example, only tag `docker-0.x.x` would trigger the publish of docker images

.github/workflows/publish-docker-images.yml
h0ngcha0: If we still want to keep maintaining `testnet`, we will need to add a new task to publish docker images for `mainnet`. Otherwise we need to update this to be `mainnet`.

h0ngcha0:
> Another important question: each tag now will trigger the publish of both docker image and SDK right?
> 
> Since the docker image would be updated less often or differently from the SDK. Maybe we use a prefix for docker image publish? For example, only tag `docker-0.x.x` would trigger the publish of docker images

Right now the SDK version is [parsed](https://github.com/alephium/wormhole-fork/blob/test-image-publish/.github/workflows/publish-sdk.yml#L25) from the package.json file. A new SDK will be published whenever a new version is detected in master.   So docker version and SDK verison are not the same atm.

Lbqds (approved):
LGTM

PR #87, opened 2023-10-17
Fix getting next governance sequence

Lbqds:
We need to get the current governance sequence from all chains and then use the `max(governanceSeqsFromAllChains) + 1` as the next governance sequence. But we cannot determine how many governance VAA have already been executed from the ETH contract(as ETH does not store sequence info but uses a map to record all executed VAA, and in Solidity, we cannot get the size of a map). So I modified it here to get the next governance sequence from the explorer-backend.


h0ngcha0 (approved):
LGTM :+1:

PR #88, opened 2023-10-19
Fix the name of wrapped token

Lbqds:
Replace ` (Wormhole)` with ` (AlphBridge)`.


polarker (approved):
LGTM, but the integration test failed.

Lbqds:
> LGTM, but the integration test failed.

Because we modified the contract, we need to release a new version of `@alephium/wormhole-sdk` and then update other packages.

PR #89, opened 2023-10-27
Improve BridgeRewardRouter contract

Lbqds:
1. Only reward bridge transfers which token is from other chains. If a user first transfers ALPH from Alephium to Ethereum and then transfers ALPH from Ethereum back to Alephium, we don't need to reward in this case.
2. Add more tests for the `BridgeRewardRouter` contract.


h0ngcha0 (approved):
LGTM :+1:

PR #96, opened 2023-11-04
Fix the blocks range limit error for ETH

Lbqds:
The log query API for ETH miannet has a maximum block range limit of 1000


h0ngcha0 (approved):
LGTM :+1:

PR #98, opened 2023-11-06
Fallback to the public node service if cannot fetch the latest block …

Lbqds:
When fetching the block number through metamask, sometimes metamask may use a cache, which causes the use of a stale block number. And I haven't found any docs about this.

In this PR, if the block number still does not change after 8 minutes(slightly longer than one epoch), it will try to get the block number from the public node RPC service.


h0ngcha0 (approved):
LGMT :+1:

PR #103, opened 2023-11-08
Fix wait tx confirmations

Lbqds:
Sometimes it throws an exception when fetching the tx status. We need to handle this exception within the function `waitALPHTxConfirmed`, otherwise it will result in not being able to retrieve the signed vaa here: https://github.com/alephium/wormhole-fork/blob/95b3d9b062635d8b31d45a41b036d85c8c322af2/bridge_ui/src/hooks/useHandleTransfer.tsx#L394


h0ngcha0 (approved):
LGTM :+1:

PR #105, opened 2023-11-14
Prepare relayer

Lbqds:


h0ngcha0:
Q: Where is the code that does the actual redeem?

Lets deploy this and test it on testnet v2, I can help with that.

Lbqds:
> Q: Where is the code that does the actual redeem?
> 
> Lets deploy this and test it on testnet v2, I can help with that.

I'm testing on the devnet now, I'll add a relayer.yml.template to the bridge infra repo.

relayer-engine/src/storage/redis-storage.ts
```diff
@@ -154,7 +154,7 @@ export class RedisStorage implements Storage {
 
   async onTxSubmitted(vaaId: string, txId: string): Promise<void> {
     const txKey = `${vaaId}:txId`.toLowerCase()
-    await this.redis.set(txKey, txId)
+    await this.redis.set(txKey, txId, 'EX', 30 * 24 * 3600)
```
polarker, line 157: Do we need 30 days cache here?

polarker (approved):
LGTM

relayer-engine/src/storage/redis-storage.ts
Lbqds: The purpose of storing the tx id is to be able to return it to the bridge ui. The assumption here is that the user will execute a redemption within 30 days, but seems that 30 days might be a bit long. How about one week?

polarker: Yes, I think one week is enough

h0ngcha0 (approved):
LGTM :+1:

PR #109, opened 2023-11-22
Improve loading alephium tokens

Lbqds:
We don't need to fetch token info from the chain, sometimes it's slow due to the rate limiter.


h0ngcha0 (approved):
LGTM :+1:

PR #114, opened 2023-12-21
Fix infinite token loading

Lbqds:
If token loading fails, it will retry indefinitely. After the fix, if token loading fails, it will pop up an error message:
Screenshot 2023-12-21 at 10 07 14

Users can reload the token list by clicking the refresh button:
Screenshot 2023-12-21 at 10 07 40


h0ngcha0 (approved):
LGTM :+1:

PR #125, opened 2024-03-14
Add banner with warning about issue being fixed

mvaivre:


bridge_ui/src/App.js
```diff
@@ -125,9 +125,33 @@ const useStyles = makeStyles((theme) => ({
...
 
+const BannerMessage = (
+  <>
+    <span>
+      We are aware of some difficulties redeeming on the bridge. The issue has
+      been identified, the fix is being worked on now. <b>Funds are safe!</b>
+    </span>
+    <i>
+      (if you want to bridge, you can bridge, but redeem will only be possible
+      once the fix is deployed)
```
polarker, line 150: Should we mention `ethereum dencun upgrade` in the message?

Lbqds (approved):
LGTM, thanks!

bridge_ui/src/App.js
mvaivre: @VDAODAO

mvaivre: I would personally mention it yes. What about:

```
We are aware of some difficulties redeeming on the bridge caused by the Ethereum Dencun upgrade. The fix is being worked on now.
```

VDAODAO: @mvaivre FYI didn't include the "because of dencun" because it wasn't 100% sure. I'd phrase it a tiny bit different: 

> We are aware of some difficulties redeeming on the bridge, following the Ethereum Dencun upgrade. 
> The fix is being worked on now.

mvaivre: So are we now sure it's the aftermath of the ethereum update? I'm AFK for an hour or so. We can merge this one and deploy rn, and I can open a new PR later this afternoon if we want to edit the text.

Lbqds: Yes, it was caused by the ETH update. ETH introduced a new transaction type, and the library used by the bridge does not support it.

mvaivre:
Apparently fixed, good job! 
But this banner could be useful in the future ;)

PR #128, opened 2024-04-15
Fetch logs by block number

Lbqds:
Currently, chainstack always returns an invalid block hash, this PR trying to fetch block logs using block height instead of hash to fix this.

### DIFF
--- a/explorer-backend/transactions/evm.go
+++ b/explorer-backend/transactions/evm.go
@@ -104,7 +104,7 @@ func (w *EVMWatcher) fetchEvents(ctx context.Context, errC chan<- error) {
 					return
 				}
 			}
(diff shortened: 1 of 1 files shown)


explorer-backend/transactions/evm.go
```diff
@@ -104,7 +104,7 @@ func (w *EVMWatcher) fetchEvents(ctx context.Context, errC chan<- error) {
 					return
 				}
 			}
-			if err := w.fetchEventsByBlockHash(ctx, block.Hash); err != nil {
+			if err := w.fetchEventsByBlockNumber(ctx, block.Number); err != nil {
 				w.logger.Error("failed to fetch events by block hash", zap.String("blockHash", block.Hash.Hex()), zap.Error(err))
```
h0ngcha0, line 108: [Nit] Shall we update this log as well?

PR #130, opened 2024-06-12
Update sdk to 1.0.0

Lbqds:


node/pkg/alephium/watcher.go
```diff
@@ -478,17 +482,29 @@ func isEventConfirmed(
...
 	return true
 }
+
+func getConfirmationDuration(
+	isMainnet bool,
+	isTransferTokenVAA bool,
+	eventConsistencyLevel uint8,
+) int64 {
+	if isMainnet && isTransferTokenVAA {
+		return int64(maxUint8(eventConsistencyLevel, MinimalConsistencyLevel)) * BlockTimeMs
```
Lbqds, line 507: Since the current `minimalConsistencyLevel` in the contract is 105, and after the Rhone upgrade, the maximum reorg size is 200, we need to wait for at least 205 * 16s. Do you think we need to update the `minimalConsistencyLevel` in the contract? This would require each guardian to sign the governance VAA.

node/pkg/alephium/watcher.go
```diff
@@ -478,17 +482,29 @@ func isEventConfirmed(
...
+	isMainnet bool,
 ) bool {
 	consistencyLevel := event.msg.consistencyLevel
 	if eventBlockHeader.Height+int32(consistencyLevel) > currentHeight {
 		logger.Debug("not enough block confirmations", zap.String("txId", event.TxId), zap.Uint8("consistencyLevel", consistencyLevel))
 		return false
 	}
 
-	duration := int64(consistencyLevel) * BlockTimeMs
+	duration := getConfirmationDuration(isMainnet, event.msg.IsTransferTokenVAA(), consistencyLevel)
```
h0ngcha0, line 493: Q: Shall we check `isMainnet` here, we probably will set up a testnet bridge again soon (to test new integrations).

Lbqds: If we update the `minimalConsistencyLevel` in the contract to 205, we won't need the `getConfirmationDuration` function anymore. We only need to wait for `consistencyLevel * 16s`, as the contract will ensure that the `consistencyLevel` is not less than 205.

h0ngcha0 (approved):
LGTM :+1: 

Will prepare commands for our Guardians to update required confirmations.

alephium/.project.json
```diff
@@ -203,7 +205,12 @@
       "sourceCodeHash": "81c5941101098365861356351470f40d88b863ddb1546e57218e3756c1f483bc",
       "bytecodeDebugPatch": "",
       "codeHashDebug": "3921ab78ca073b5125b2ff7e4232e72f2c1c62554bbc8f81d39e769e6bcf349c",
-      "warnings": []
+      "warnings": [
```
polarker, line 208: Side note (not for this PR): In web3 SDK, we could remove warnings when saving the file.

polarker (approved):
LGTM

PR #136, opened 2024-07-02
Fix the eth token list url

Lbqds:
The ETH token list URL we use has changed several times recently, and every time the URL changes, using the old URL will cause CORS errors in the browser, and we have to manually update the code to the new URL and redeploy the bridge ui. Do you think it is necessary for us to maintain an ETH token list on Github?


h0ngcha0 (approved):
LGTM :+1:

PR #139, opened 2024-08-21
Fix checking token info

Lbqds:


bridge_ui/src/utils/alephium.ts
```diff
@@ -58,7 +58,7 @@ async function fetchTokenList(): Promise<TokenList> {
   return tokenList
 }
 
-async function getTokenFromTokenList(tokenId: string): Promise<TokenInfo | undefined> {
+async function getTokenFromTokenList(tokenId: string): Promise<TokenInfo & { nameOnChain?: string; symbolOnChain?: string } | undefined> {
```
h0ngcha0, line 61: [Nit] Maybe we should fix the `TokenInfo` type?

h0ngcha0 (approved):
LGTM :+1:

bridge_ui/src/utils/alephium.ts
Lbqds: The `TokenInfo` type is from `token-list` repo: https://github.com/alephium/token-list/blob/a2fed941bd98c91bffea257879a075d7df7ac95c/lib/types.ts#L24, I remember these two fields were in the previous version, but they were removed in the latest version.

h0ngcha0: IIRC these two fields were introduced recently because of the name/symbol conflicts of name services, but we probably didn't update the `TokenInfo` type.

Lbqds: Yeah, these two fields were introduced in v0.0.12: https://github.com/alephium/token-list/blob/v0.0.12/lib/types.ts, but they were removed later.

h0ngcha0: I see, thanks. I am not sure about the reasoning behind removing it, maybe we should introduce it back? 
In any case maybe we should merge this PR so ANS can register token :+1:

Lbqds: This is the PR that removed these two fields: https://github.com/alephium/token-list/pull/59

PR #146, opened 2024-11-04
Fix token decimals from BSC

Lbqds:
The decimals for USDT and USDC on BSC are 18, which is different from ETH.


polarker:
Maybe let's add a test for mainnet metadata?

Lbqds:
> Maybe let's add a test for mainnet metadata?

Sounds great, will add!

polarker (approved):
LGTM

PR #147, opened 2024-11-05
Overlay the original token logo and symbol for bridged tokens

Lbqds:
The source chain is Alephium and the target chain is Ethereum:
Screenshot 2024-11-06 at 20 37 56

The source chain is Alephium and the target chain is BSC
Screenshot 2024-11-06 at 20 38 03


bridge_ui/src/utils/evm.ts
```diff
@@ -55,14 +87,15 @@ export async function getEthereumToken(
 }
 
 export async function ethTokenToParsedTokenAccount(
```
h0ngcha0, line 89: [Nit] `evmTokenToParsedTokenAccount`

h0ngcha0 (approved):
LGTM, only one minor comment.

polarker:
Q: Should we filter the tokens based on the network connected? E.g. If I am connected to Ethereum, then only eth tokens should be selectable

Lbqds:
> Q: Should we filter the tokens based on the network connected? E.g. If I am connected to Ethereum, then only eth tokens should be selectable

Sounds great 👍, will update

polarker (approved):
LGTM

PR #150, opened 2024-11-20
Fix the wormhole sdk in react native env

Lbqds:
There is some WASM code in our wormhole sdk, which is not very compatible with the react native environment. Additionally, it also uses some nodejs libraries, which further complicate compatibility.

In the latest official wormhole sdk, typescript code has been used instead of WASM, and the nodejs libraries have been removed. So I removed the Solana/Terra/Algorand-related code from our sdk in this PR, making it easier to maintain.


polarker (approved):
LGTM

PR #151, opened 2024-12-25
Fix walletconnect issues

Lbqds:


bridge_ui/package.json
```diff
@@ -23,7 +23,7 @@
     "@solana/web3.js": "^1.35.1",
     "@terra-money/wallet-provider": "^3.8.0",
     "@alephium/walletconnect-qrcode-modal": "0.1.0",
-    "@alephium/walletconnect-ethereum-provider": "0.1.1",
+    "@walletconnect/ethereum-provider": "2.17.2",
```
Lbqds, line 26: In the latest WalletConnect, we will separate the WalletConnect storage for Alephium and Ethereum. So we no longer need to maintain a custom Ethereum provider.

bridge_ui/src/contexts/EthereumProviderContext.tsx
```diff
@@ -219,8 +221,9 @@ export const EthereumProviderProvider = ({
         EthereumProvider.init({
           projectId: WALLET_CONNECT_PROJECT_ID,
           showQrModal: false,
-          chains: [getEvmChainId(wormholeChainId) as number],
-          rpcMap: EVM_RPC_MAP
+          chains: [getEvmChainId(CHAIN_ID_ETH) as number, getEvmChainId(CHAIN_ID_BSC) as number],
```
Lbqds, line 224: In the latest WalletConnect, switching chains does not require user approval but instead requires specifying all chain ids when initializing.

polarker (approved):
LGTM 👍

bridge_ui/src/index.js
```diff
@@ -15,9 +15,14 @@ import { TerraWalletProvider } from "./contexts/TerraWalletContext.tsx";
 import ErrorBoundary from "./ErrorBoundary";
 import { theme } from "./muiTheme";
 import { store } from "./store";
-import { AlephiumWalletProvider } from "@alephium/web3-react";
+import { AlephiumWalletProvider, createWalletConnectConnector, createDesktopWalletConnector } from "@alephium/web3-react";
 import { CLUSTER, ALEPHIUM_BRIDGE_GROUP_INDEX } from "./utils/consts";
 
+const connectors = {
+  walletConnect: createWalletConnectConnector({customStoragePrefix: 'alephium'}),
+  desktopWallet: createDesktopWalletConnector({customStoragePrefix: 'alephium'})
```
polarker, line 23: Maybe we should enable this by default in our walletconnect provider? There could be Other projects need to use multiple connectors.

Lbqds: Sounds good, will update

PR #155, opened 2025-03-14
Fix eth full node url

Lbqds:


bridge_ui/src/hooks/useCheckIfWormholeWrapped.ts
```diff
@@ -101,7 +102,7 @@ function useCheckIfWormholeWrapped(nft?: boolean) {
     (async () => {
       if (isEVMChain(sourceChain) && sourceAsset) {
         // use JsonRpcProvider if both source chain and target chain are evm chains
-        const provider = evmChainId === sourceChain ? evmProvider : getEvmJsonRpcProvider(sourceChain)
+        const provider = evmChainId === getEvmChainId(sourceChain) ? evmProvider : getEvmJsonRpcProvider(sourceChain)
```
Lbqds, line 105: We should convert the wormhole chain id to the EVM chain id before comparing. I checked the entire codebase; this is the only place with an issue.

bridge_ui/src/utils/consts.ts
```diff
@@ -276,16 +276,16 @@ export const SOLANA_HOST = process.env.REACT_APP_SOLANA_API_URL
 
 export const ETH_RPC_HOST =
   CLUSTER === 'mainnet'
-    ? 'https://ethereum-rpc.publicnode.com'
+    ? 'https://eth-mainnet.public.blastapi.io'
```
Lbqds, line 279: The full node service provided by the public node is down, and it seems that their website is also inaccessible. So I have switched the backup full node URL to blast API: https://blastapi.io/. Let me know if it's ok.

h0ngcha0 (approved):
LGTM :+1:

polarker (approved):
LGTM!

PR #157, opened 2025-05-08
Fix the issue of fetching the token list

Lbqds:
1. First, try to fetch the token whitelist from the original token URL.
2. If that fails, fall back to fetching it from github.

If we need to update the whitelist, we only need to update the github repo without redeploying the bridge ui.


polarker:
Maybe we could load the JSON file directly in TS instead of downloading it from Github?

Lbqds:
> Maybe we could load the JSON file directly in TS instead of downloading it from Github?

The question is whether we need to update the token list. If so, we have to redeploy the bridge ui every time we make an update.

polarker:
> > Maybe we could load the JSON file directly in TS instead of downloading it from Github?
> 
> The question is whether we need to update the token list. If so, we have to redeploy the bridge ui every time we make an update.

I think there's no need to update that often, it's stable 99% of the time. For a better UX, I'll just load the JSON file in the code instead of fetching it from the web.

h0ngcha0 (approved):
LGTM :+1:

PR #169, opened 2025-10-02
Bridge redesign - 3

mvaivre:
- Redesigned the registration (attest flow)
- Manage wallet disconnection during the flow
- Misc improvements accros bridge widget


mvaivre:
>I see that you editted a lot of existing code that we consider "Legacy tools" which makes me a bit unconfortable in the event that we break something that used to work. With the new "bridge widget" UI we did not touch the existing "Transfer" flow of the legacy tools. But this PR changes a lot of logic of the token registration which is accessible from both the bridge widget and the legacy tools transfer flow.

I can create a duplicate, no problem. New UI would use the redesigned flow, legacy UI would redirect to the legacy register flow.

PR #178, opened 2025-10-22
Bridge redesign feedback fixes

nop33:
1. Updates code to support groupless addresses
2. Fixes the display of the block finality state when user navigates to a different page of the app
4. Fixes broken final state on transfer completion
5. Improves display of remaining seconds for block confirmation
6. Puts chain-swap button in the center (only for desktop)
7. Fixes Alephium connect button in /bridge page and in /register page
8. Displays estimated Alephium block finality time
9. Updates SDK to make node & explorer backend requests within the dApp instead from the wallet through WC

### DIFF
--- a/bridge_ui/package-lock.json
+++ b/bridge_ui/package-lock.json
@@ -10,8 +10,8 @@
       "dependencies": {
         "@alephium/token-list": "0.0.12",
         "@alephium/walletconnect-qrcode-modal": "0.1.0",
-        "@alephium/web3": "^2.0.4",
-        "@alephium/web3-react": "^2.0.4",
+        "@alephium/web3": "^2.0.7",
+        "@alephium/web3-react": "^2.0.7",
         "@alephium/wormhole-sdk": "0.6.0-rc.1",
         "@material-ui/core": "^4.12.2",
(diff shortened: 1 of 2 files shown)


bridge_ui/src/components/BridgeWidget/MainActionButton.tsx
```diff
@@ -27,7 +27,6 @@ interface MainActionButtonProps {
 const MainActionButton = ({ onNext }: MainActionButtonProps) => {
   const classes = useStyles()
   const { t } = useTranslation()
-  const { connect: connectAlephium } = useConnect()
```
nop33: Even though our docs say that we can use `useConnect` to replace the `AlephiumConnectButton`, in reality the `connect` function returned by `useConnect` does not show the modal where the user can select which wallet they want to connect to (WalletConnect / Extension / DW), but instead it chooses the extension wallet. I couldn't connect my mobile wallet.

nop33:
@mvaivre I'll merge this so that the blockchain team can already review our feature branch PR (our focus should move to the core dApp so I am trying to push the bridge to be done). Please, feel free to review this PR as we normally do. I can apply any suggested changes in a follow-up.

PR #185, opened 2025-10-29
Redesign feedback

nop33:
image


mvaivre (approved):
Apart from the tiny padding nit shared on Slack, LGTM! Approving already.

PR #187, opened 2025-10-30
Update walletconnect ethereum provider

Lbqds:
We need to merge this and release a new version: https://github.com/alephium/walletconnect-ethereum-provider/pull/2

### DIFF
--- a/bridge_ui/package-lock.json
+++ b/bridge_ui/package-lock.json
@@ -9,6 +9,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@alephium/token-list": "0.0.12",
+        "@alephium/walletconnect-ethereum-provider": "0.2.0",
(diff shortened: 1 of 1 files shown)


bridge_ui/src/contexts/EthereumProviderContext.tsx
```diff
@@ -222,7 +221,6 @@ export const EthereumProviderProvider = ({
           projectId: WALLET_CONNECT_PROJECT_ID,
           showQrModal: false,
           chains: [getEvmChainId(CHAIN_ID_ETH) as number, getEvmChainId(CHAIN_ID_BSC) as number],
-          rpcMap: EVM_RPC_MAP,
```
Lbqds: We no longer need to set the `rpcMap` because WalletConnect has a default RPC URL for each chain. Also, the RPC URLs in the bridge UI no longer work.

PR #188, opened 2025-11-03
Integrate recovery and tx history in new UI

### LINKED ISSUE #183: Integrate Redeem page and Transactions page into the new bridge UI

### LINKED ISSUE #184: Inform user that they can copy the tx hash to redeem later if they leave the page

nop33:
- Closes #183 
- Closes #184


bridge_ui/src/components/BridgeWidget/ChainSelectors.tsx
nop33: Moved the chain selectors into its own component so that it can be reused in 2 places:
- enter data step of bridging
- tx history page

bridge_ui/src/components/BridgeWidget/SendingAddress.tsx
nop33: In the recover flow we cannot recover the hash of the sending (source) address, but we know the chain. So I decided to update the UI to display "from an Alephium address" instead of saying "From Alephium address [missing address]"

bridge_ui/src/components/BridgeWidget/SmartAddress.tsx
nop33: To solve the issue where we need to inform the user that they can copy the tx id in order to recover their transfer in case they navigate away from the page, I added a pulsing pointer emoji animation which gets hidden the first time the user hovers over or clicks the tx id.
- https://github.com/alephium/wormhole-fork/issues/184

bridge_ui/src/components/BridgeWidget/TransactionsHistory/TransactionsHistory.tsx
nop33: Same as above

mvaivre:
https://github.com/alephium/wormhole-fork/pull/189

nop33:
@mvaivre Thanks for your suggestions and implementation! I created a follow-up PR to fix a bug and some misalignments that I noticed:
- #190

PR #190, opened 2025-11-04
Fix url query params reset

nop33:
@mvaivre there was a bug in your PR where the URL query params where not reset when navigating away from the Recovery page, which is confusing for the user. In this PR I simplify the code and I reset the query params when they are not needed anymore. I understand that you utilized the query params so that you can transfer the tx id when redirect to the Recovery page from the History page.

### DIFF
--- a/bridge_ui/src/components/BridgeWidget/TransactionTableCompact.tsx
+++ b/bridge_ui/src/components/BridgeWidget/TransactionTableCompact.tsx
@@ -42,15 +42,14 @@ const useStyles = makeStyles(() => ({
     display: 'inline-flex',
     alignItems: 'center',
     justifyContent: 'center',
-    width: 18,
-    height: 18,
-    borderRadius: '50%'
+    width: 14,
+    height: 14,
   },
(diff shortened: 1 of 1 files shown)


bridge_ui/src/components/BridgeWidget/EnterDataStep/BridgeWidgetNavItem.tsx
nop33: just some cleanup, keeping code DRY

bridge_ui/src/components/BridgeWidget/TransactionTableCompact.tsx
```diff
@@ -44,14 +44,13 @@ const useStyles = makeStyles(() => ({
     justifyContent: 'center',
     width: 18,
     height: 18,
-    borderRadius: '50%',
-    marginRight: 8
```
nop33: I changed the position of the icon cause it looked misaligned

| before | after |
| ----- | ----- |
| image | image |

PR #193, opened 2025-11-10
Upgrade Material UI packages from v4 to v7

nop33:
I incrementally upgraded the Material UI packages from v4 to v5 to v6 and finally to v7, so that I can later on create an NPM package with modern build tools like Vite, similarly to how Wormhole Connect is doing it.

My next goal is to upgrade React from 17 to 19. But I'll create a separate PR for that.

@mvaivre, I tried hard to fix any styling inconsistencies. If you notice anything wrong, please help me fix it or ping me.


mvaivre (approved):
LGTM!

PR #196, opened 2025-11-10
Remove react-router dependency in widget code

nop33:
- Moves token registration flow into the widget and out of react-router
- Cleans up components and groups them in a logical folder structure
- Removes dependency on react-router when navigating from tx history page to recovery page


nop33:
These styling issues can be addressed later. Created issue:
- https://github.com/alephium/wormhole-fork/issues/199

mvaivre (approved):

